PR #67, opened 2021-07-21
Add support for nested paths update

NouemanKHAL:


tests/integration/helpers.py
```diff
@@ -60,7 +60,14 @@ def test_resource_update_sync(self, script_runner):
...
         for resource in source_resources.values():
-            resource[self.field_to_update] = str(resource[self.field_to_update]) + "+ updated"
+            try:
+                current_value = pathLookup(resource, self.field_to_update)
+                if current_value is None:
+                    current_value = ""
+
+                pathUpdate(resource, self.field_to_update, current_value + "Updated")
+            except Exception as e:
+                print("test_resource_update_sync error: " + str(e))
```
skarimo, line 70: Should this fail instead if we cannot update the path?

NouemanKHAL: You're absolutely right, I did this for debugging purposes.

skarimo (approved):

PR #69, opened 2021-07-27
Use tox to manage test environment

NouemanKHAL:

PR #70, opened 2021-07-27
Use RECORD env var to control record mode

NouemanKHAL:
- `RECORD=true` for recording, `RECORD=false` for replaying and `RECORD=none` for using the real API without recording
- Default values for environment vars will be taken care of in #69


tests/conftest.py
```diff
@@ -41,9 +41,18 @@ def before_record_response(response):
     return before_record_response
 
 
+def get_record_mode():
+    record = os.getenv("RECORD", "none")
+    if record.lower() == "true":
+        return "all"
+    else:
+        return "none"
```
zippolyte, line 49: There are three cases `RECORD=true` for recording, `RECORD=false` for replaying and `RECORD=none` for using the real API without recording

zippolyte (changes requested):

tests/conftest.py
NouemanKHAL: I don't think there's an option to not record : https://vcrpy.readthedocs.io/en/latest/usage.html#record-modes
Unless you pass the --disable-recording flag

zippolyte: We have these 3 modes in the API clients, so let's have them here too

.github/workflows/test.yml
```diff
@@ -35,10 +35,10 @@ jobs:
             - name: Install datadog-sync-cli
               run: pip install .\[tests\]
             - name: Run unit tests
-              run: pytest --record-mode=none -ra -v
+              run: pytest -ra -v
             - name: Run integration tests
               if: contains(github.event.pull_request.labels.*.name, 'ci/integration')
-              run: pytest --disable-recording -ra -v -m "integration"
+              run: RECORD=true pytest -ra -v -m "integration"
```
jirikuncar, line 41:
```suggestion
              run: RECORD=none pytest -ra -v -m "integration"
```

zippolyte (approved):

PR #73, opened 2021-07-29
CLI tool redesign

skarimo:


datadog_sync/model/dashboard_lists.py
```diff
@@ -4,115 +4,103 @@
...
-        "root['created']",
-        "root['modified']",
-        "root['is_favorite']",
-        "root['dashboard_count']",
-    ]
-
-    def import_resources(self):
-        source_client = self.config.source_client
 
+    def get_resources(self, client: CustomClient) -> List[Dict]:
```
skarimo, line 26: Still deciding if it's worth moving this method into the base resource as it is relatively similar in each resource. Exceptions being return field used + whether the endpoint is paginated or not.

skarimo: Also, this PR is getting huge so few changes will follow in a follow up PR:
- global exception handler so we don't have to `try` and `catch` each request within resources
- Move the action of adding/updating resources in the resource objects to the base resource e.g. https://github.com/DataDog/datadog-sync-cli/blob/299a9cb4cd799f72b40f8a67e79dcd73d6852080/datadog_sync/model/dashboard_lists.py#L49

therve (approved):

datadog_sync/utils/base_resource.py
```diff
@@ -3,190 +3,207 @@
...
+
+        futures = []
+        with thread_pool_executor(max_workers) as executor:
+            if resources_list:
+                for r_list in resources_list:
+                    for _id, resource in r_list.items():
+                        if not self.filter(resource):
+                            continue
+                        futures.append(executor.submit(self.apply_resource, _id, resource))
+                    wait(futures)
```
NouemanKHAL, line 134: why do we wait for the futures here?

datadog_sync/model/monitors.py
```diff
@@ -4,99 +4,74 @@
...
 
-        for _id, monitor in self.source_resources.items():
-            if monitor["type"] == "synthetics alert":
-                continue
+        for _id, monitor in self.resource_config.source_resources.items():
             if monitor["type"] == "composite":
                 composite_monitors[_id] = monitor
             else:
                 simple_monitors[_id] = monitor
+        return [simple_monitors, composite_monitors]
```
NouemanKHAL, line 57:
```suggestion
        return simple_monitors + composite_monitors
```
We can merge them and keep the list flat, right?

datadog_sync/utils/base_resource.py
```diff
@@ -3,190 +3,207 @@
...
+        try:
+            resources_list = self.pre_apply_hook(self.resource_config.source_resources)
+        except Exception as e:
+            self.config.logger.error(f"error while applying resources {self.resource_type}: {str(e)}")
+            return
+
+        futures = []
+        with thread_pool_executor(max_workers) as executor:
+            if resources_list:
+                for r_list in resources_list:
```
NouemanKHAL, line 129: if we keep the list flat, we won't need the second for loop here.

datadog_sync/utils/base_resource.py
```diff
@@ -3,190 +3,207 @@
...
+                    for _id, resource in r_list.items():
+                        if not self.filter(resource):
+                            continue
+                        futures.append(executor.submit(self.apply_resource, _id, resource))
+                    wait(futures)
+            else:
+                for _id, resource in self.resource_config.source_resources.items():
+                    if not self.filter(resource):
+                        continue
+                    futures.append(executor.submit(self.apply_resource, _id, resource))
```
NouemanKHAL, line 139:
```suggestion
            if not resources_list:
                resources_list = self.resource_config.source_resources
        
            for r_list in resources_list:
                for _id, resource in r_list.items():
                    if not self.filter(resource):
                        continue
                    futures.append(executor.submit(self.apply_resource, _id, resource))
```
Might be better this way, less duplicated code

datadog_sync/utils/base_resource.py
skarimo: We wait to make sure they are finished prior to moving on to the next object in the list. To ensure order is kept and there are no cases where we move on to the next item on the list before the current one finishes.

This is pretty much doing the same thing you implemented in the monitors resource originally but without having to call apply manually, twice, as in the original case.

datadog_sync/model/monitors.py
skarimo: No we cannot. Combining the list would mean that there is a possibility that composite monitor is applied at the same time as simple monitors.

This goes back to your original approach as well. Pretty much the same concept but without having to call apply manually twice.

datadog_sync/utils/base_resource.py
skarimo: Check the reply above.

datadog_sync/utils/base_resource.py
skarimo: Same here, check the reply above. This is a workaround to ensure order is kept. E.g. simple and composite monitors

NouemanKHAL (approved):
LGTM!

PR #76, opened 2021-08-03
Remove unused cassettes

NouemanKHAL:
Remove old [inprocess] cassettes since we no longer use `script_runner`.


skarimo (changes requested):
Can you remove the ones in integration/resources directory as well

skarimo (approved):

PR #79, opened 2021-09-08
Use custom exception class to format HTTP Errors

skarimo:


datadog_sync/utils/resource_utils.py
```diff
@@ -24,6 +24,11 @@ def __init__(self, resource_type, _id=None):
         )
 
 
+class CustomClientHTTPError(Exception):
+    def __init__(self, response):
+        super(CustomClientHTTPError, self).__init__(f"{response.status_code} {response.reason} - {response.text}")
```
NouemanKHAL, line 29:
```suggestion
        super().__init__(f"{response.status_code} {response.reason} - {response.text}")
```
Just a nit, in python3 it's no longer necessary to pass the args

NouemanKHAL (approved):

PR #82, opened 2021-09-22
Handle secret synthetics global variables 

NouemanKHAL:
Add default"SECRET" dummy value to secret synthetics global variables to fix syncing issues : `["Required parameter 'value' is missing"]`

### DIFF
--- a/datadog_sync/model/synthetics_global_variables.py
+++ b/datadog_sync/model/synthetics_global_variables.py
@@ -35,6 +35,12 @@ class SyntheticsGlobalVariables(BaseResource):
     def get_resources(self, client: CustomClient) -> List[Dict]:
         resp = client.get(self.resource_config.base_path).json()
 
+        # case of secret variable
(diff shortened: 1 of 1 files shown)


datadog_sync/model/synthetics_global_variables.py
```diff
@@ -32,9 +32,19 @@ class SyntheticsGlobalVariables(BaseResource):
     # Additional SyntheticsGlobalVariables specific attributes
     destination_global_variables: Dict[str, Dict] = dict()
 
+    def pre_process(self, resource: Dict) -> None:
```
skarimo, line 35: Are you able to use the `pre_resource_action_hook` method instead to handle this since it runs before applying?

skarimo: Also, looks like `value` is only needed on create so it might be better to just handle it in the create step so we do not override values on subsequent updates.

skarimo (approved):

PR #84, opened 2021-10-05
Skip slo alert for monitors and fix default resource connections for type list

skarimo:
This pr contains 2 fixes
1) Skips monitor  type `slo alert` as it is not created directly via montiors endpoint
2) Fixes default resource connection when attribute is list of id's


skarimo:
Thanks for pointing this out. Will be fixed by: https://github.com/DataDog/datadog-sync-cli/pull/149

PR #87, opened 2021-12-20
Few resource fixes

skarimo:
This PR:

- adds ability to customize pagination
- fixes resource connections in few resources (monitors,)
- temporary fix for global variables stored in synthetics_tests

### DIFF
--- a/datadog_sync/model/monitors.py
+++ b/datadog_sync/model/monitors.py
@@ -91,3 +91,6 @@ class Monitors(BaseResource):
                 if not found:
                     raise ResourceConnectionError(resource_to_connect, _id=_id)
             r_obj[key] = (r_obj[key].replace("#", "")).strip()
+        elif key != "query":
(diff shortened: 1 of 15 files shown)


datadog_sync/model/monitors.py
```diff
@@ -91,3 +91,6 @@ def connect_id(self, key: str, r_obj: Dict, resource_to_connect: str) -> None:
                 if not found:
                     raise ResourceConnectionError(resource_to_connect, _id=_id)
             r_obj[key] = (r_obj[key].replace("#", "")).strip()
+        elif key != "query":
+            # Use default connect_id method in base class when not haandling special case for `query`
```
NouemanKHAL, line 95:
```suggestion
            # Use default connect_id method in base class when not handling special case for `query`
```

therve (approved):

NouemanKHAL (approved):

NouemanKHAL (approved):

PR #91, opened 2022-02-25
Fix tests

skarimo:
- Update CI to use tox
- remove unneeded dependencies
- Fix tox 
- Disable some tests from running in non-replay mode

Some tests are skipped until I introduce proper resource cleanup.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,3 @@
 [pytest]
 markers =
     integration: marks integration tests that require source/destination api+app keys.
-env =
-    MAX_WORKERS=1
--- a/tests/integration/resources/test_monitors.py
+++ b/tests/integration/resources/test_monitors.py
@@ -10,7 +10,7 @@ from datadog_sync.models import Monitors
 
 
 @pytest.mark.skipif(
-    get_record_mode() != "none", reason="Cannot create duplicate monitors. Reintroduce with  proper cleanup."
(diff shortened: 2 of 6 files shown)


tests/integration/resources/test_monitors.py
```diff
@@ -2,11 +2,16 @@
...
 # Copyright 2019 Datadog, Inc.
+import pytest
 
+from tests.conftest import get_record_mode
 from tests.integration.helpers import BaseResourcesTestClass
 from datadog_sync.models import Monitors
 
 
+@pytest.mark.skipif(
+    get_record_mode() != "none", reason="Cannot create duplicate monitors. Reintroduce with  proper cleanup."
```
NouemanKHAL, line 13:
```suggestion
    get_record_mode() != "none", reason="Cannot create duplicate monitors. Reintroduce with proper cleanup."
```

pytest.ini
```diff
@@ -1,5 +1,3 @@
 [pytest]
 markers =
     integration: marks integration tests that require source/destination api+app keys.
-env =
-    MAX_WORKERS=1
```
NouemanKHAL, line 5: Don't we need this for `pytest-recording` ?

skarimo: Using tox now to set this and run tests so no need for this extra setting + dep

NouemanKHAL: Yes, just thought maybe we can keep it for safety, it's up to you.

skarimo: Ill omit for now and. reintroduce if needed later on

PR #92, opened 2022-04-15
Add Logs indexes and Logs restriction query resources

skarimo:
Add 2 new resources:
- logs_indexes
- logs_restriction_queries

Few small fixes are included here as well:
1) avoid unnecessary diff in downtime when the start and end times are different between source and destination.
2) Slo correction with `end` time > 90 days in the past are not synced (rejected by backend validations if we try)


datadog_sync/model/downtimes.py
```diff
@@ -24,7 +24,7 @@ class Downtimes(BaseResource):
         resource_connections={"monitors": ["monitor_id"]},
         non_nullable_attr=["recurrence.until_date", "recurrence.until_occurrences"],
         base_path="/api/v1/downtime",
-        excluded_attributes=["id", "updater_id", "created", "org_id", "modified", "creator_id", "active"],
+        excluded_attributes=["id", "updater_id", "created", "org_id", "modified", "creator_id", "active", "start"],
```
skarimo, line 27: Was testing something out here. Will revert this change

datadog_sync/utils/custom_client.py
```diff
@@ -113,6 +113,9 @@ def wrapper(*args, **kwargs):
 
                 resp_json = resp.json()
                 resources.extend(resp_json["data"])
+                if len(resp_json["data"]) < page_size:
```
NouemanKHAL, line 116:
```suggestion
                if len(resp_json["data"]) <= page_size:
```
Right?

therve (approved):

PR #99, opened 2022-07-25
Add support for `AND` filter operator

skarimo:
SSIA


datadog_sync/commands/shared/options.py
```diff
@@ -85,7 +85,14 @@
         type=int,
         help="Max number of workers when running operations in multi-threads.",
     ),
-    option("--filter", required=False, help="Filter resources.", multiple=True),
+    option(
+        "--filter-operator",
+        envvar=constants.DD_FILTER_OPERATOR,
```
skarimo, line 90: Open to name suggestions instead of calling it operator. Kind of clashes with: https://github.com/DataDog/datadog-sync-cli#filtering

therve (approved):

PR #115, opened 2023-03-17
Add formal sync order

skarimo:


therve (dismissed):
I guess the AWS sync never worked?
On the surface it seems fine, it's a bit big to completely grasp what's going on to be honest. I guess we can rely on testing to make sure things are fine.

skarimo:
> I guess the AWS sync never worked? On the surface it seems fine, it's a bit big to completely grasp what's going on to be honest. I guess we can rely on testing to make sure things are fine.

Yes it never worked. The boiler/dead code was there but never added officially as a supported resource

PR #233, opened 2024-04-23
[APITL-862] Add support for moving index to end of order list during deletion

skarimo:


luchenyuxx (approved):

luchenyuxx (commented):
Is it possible to add an unit test for this?

luchenyuxx (approved):

tests/integration/resources/test_logs_indexes.py
```diff
@@ -3,16 +3,54 @@
...
+                "--cleanup=force",
+            ],
+        )
+        assert 0 == ret.exit_code
+
+        api_key = os.environ.get("DD_DESTINATION_API_KEY")
+        app_key = os.environ.get("DD_DESTINATION_APP_KEY")
+        api_url = os.environ.get("DD_DESTINATION_API_URL")
+
+        req = urllib.request.Request(
```
tim-chaplin-dd, line 48: I think to prove that the order changed you also need to call this prior to `sync` and assert that `first_index` comes first in the ordering.

tests/integration/resources/test_logs_indexes.py
```diff
@@ -3,16 +3,54 @@
...
-    @pytest.mark.skip(reason="delete is not supported for logs-indexes")
-    def test_resource_cleanup(self):
-        pass
+    def test_resource_cleanup(self, runner, caplog):
+        caplog.set_level(logging.DEBUG)
+
+        source_resources, _ = open_resources(self.resource_type)
+
+        # Remove the first resource from the source
+        first_index = list(source_resources.keys())[0]
```
tim-chaplin-dd, line 27: Where does the data for this test come from? It'd be easier to visually verify that it's correct if it was hardcoded in the test itself. I think the test trivially passes if `source_resources` is length 1, or if all of its elements are identical, and it's not clear whether that's the case here.

tim-chaplin-dd (changes requested):

tests/integration/resources/test_logs_indexes.py
tim-chaplin-dd: Nevermind, Sherzod explained the cassette thing to me offline, I don't think this comment is relevant.

PR #237, opened 2024-04-24
[APITL-855] Add support for enabling logs integration pipelines

skarimo:
Logs integration pipeline are created by submitting a log message with a supported `source`.  This PR adds support for submitting a log to destination logs intake during create step ensuring the pipeline is enabled for the destination org.


datadog_sync/model/logs_pipelines.py
```diff
@@ -48,18 +50,42 @@ async def pre_apply_hook(self) -> None:
...
+                payload = {
+                    "ddsource": source,
+                    "ddtags": ",".join(DEFAULT_TAGS),
+                    "message": f"[datadog-sync-cli] Triggering creation of '{resource['name']}' integration pipeline",
+                }
+
+                # Submit a log to the logs intake API to trigger the creation of the integration pipeline
+                await destination_client.post(self.logs_intake_path, payload, subdomain=self.logs_intake_subdomain)
+                created = False
+                for _ in range(3):
```
luchenyuxx, line 70: NIT: I think it's probably safer to wait longer on this, it can take time to ingest this log, I'm thinking that we can make it 1 minutes (12 * 5 seconds)

datadog_sync/model/logs_pipelines.py
```diff
@@ -48,18 +50,42 @@ async def pre_apply_hook(self) -> None:
...
+                    updated_pipelines = await self.get_destination_integration_pipelines()
+                    if resource["name"] in updated_pipelines:
+                        self.destination_integration_pipelines = updated_pipelines
+                        created = True
+                        break
+                    else:
+                        await sleep(5)
+
+                if not created:
+                    raise Exception(f"Max retry reached. Integration pipeline '{resource['name']}' pending creation")
```
luchenyuxx, line 80: NIT: We are not retrying the creation of integration pipeline, but retrying to verify its creation. Maybe change the message to avoid confusion
```suggestion
                    raise Exception(f"Integration pipeline '{resource['name']}' is not created after x seconds. It will be rechecked in the next sync.")
```

datadog_sync/model/logs_pipelines.py
```diff
@@ -48,18 +50,42 @@ async def pre_apply_hook(self) -> None:
...
+
             self.resource_config.destination_resources[_id] = self.destination_integration_pipelines[resource["name"]]
-            return await self.update_resource(_id, resource)
+            diff = check_diff(self.resource_config, self.destination_integration_pipelines[resource["name"]], resource)
+            if diff:
+                return await self.update_resource(_id, resource)
```
luchenyuxx, line 85: I wonder why we try to update resource here? this is the creation endpoint, right?

BTW, there is no need to update integration pipeline at all. There are read only.

skarimo: The update is to sync `enables/disables` status of the integration pipelines

luchenyuxx: 👍 Ah, ok, I get it now. Thanks for the explanation! So, probably it's better to add a comment here in the code about the intention of the update.

luchenyuxx (approved):

tim-chaplin-dd (approved):

PR #238, opened 2024-04-24
[APITL-856] Add support for logs-indexes-order

skarimo:


buraizu (dismissed):

datadog_sync/model/logs_indexes_order.py
```diff
@@ -0,0 +1,93 @@
...
+        resource["index_names"] = [_id for _id in resource["index_names"] if _id not in ids_to_omit]
+        resource["index_names"] = resource["index_names"] + extra_ids_to_include
+
+        destination_client = self.config.destination_client
+        resp = await destination_client.put(self.resource_config.base_path, resource)
+
+        return _id, resp
+
+    async def delete_resource(self, _id: str) -> None:
+        self.config.logger.warning("logs_indexes_order cannot deleted. Removing resource from config only.")
```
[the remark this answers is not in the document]
skarimo: 👋 the "config" in this case was referring to the locally stored state files. I opted to remove the log message entirely. Thanks

luchenyuxx (dismissed):

datadog_sync/model/logs_indexes_order.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self.resource_config.destination_resources[_id] = self.destination_indexes_order
+        return await self.update_resource(_id, resource)
+
+    async def update_resource(self, _id: str, resource: Dict) -> Tuple[str, Dict]:
+        destination_resources = self.destination_indexes_order or self.resource_config.destination_resources[_id]
+        ids_to_omit = set(resource["index_names"]) - set(destination_resources["index_names"])
+
+        extra_ids_to_include = [
+            _id for _id in destination_resources["index_names"] if _id not in resource["index_names"]
```
tim-chaplin-dd, line 63: nit: pick a different variable name here and below since `_id` is also a function parameter?

datadog_sync/model/logs_indexes_order.py
```diff
@@ -0,0 +1,93 @@
...
+        self.destination_indexes_order = await self.get_destination_indexes_order()
+
+    async def create_resource(self, _id: str, resource: Dict) -> Tuple[str, Dict]:
+        if not self.destination_indexes_order:
+            raise Exception("Failed to retrieve destination orgs logs index order")
+
+        self.resource_config.destination_resources[_id] = self.destination_indexes_order
+        return await self.update_resource(_id, resource)
+
+    async def update_resource(self, _id: str, resource: Dict) -> Tuple[str, Dict]:
```
tim-chaplin-dd, line 58: What's the intended behavior of this function? IMO it's not clear why we omit some ids or consider some extra. Comments would help, but even clearer would be factoring out the set arithmetic into its own function and writing some unit tests for it.

tim-chaplin-dd (changes requested):

datadog_sync/model/logs_indexes_order.py
skarimo: Thanks I will add a test. In regards to why, this is the tricky part of dealing with logs indexes order. The main use case of both `ids_to_omit` and `extra_ids_to_include` is to account for the potential logs index differences between R1 and R2. For example given:
```
org1 indexes: [alt, main, non-existent]
org2 indexes: [main, alt, new]
```
For order syncing purposes we need the order of org1 and the intersection of the two orgs + all of the extra indexes in org2. So the resulting payload needs to be: `[alt, main, new]`

tim-chaplin-dd (approved):

PR #246, opened 2024-05-15
Ensure only valid indexes are imported for `logs_indexes_order`

skarimo:
APITL-891

This ensures that deleted log indexes are moved to the end of the index order during syncing.

`/config/index-order` endpoint returns all log indexes, including those pending full deletion. We need to make sure we exclude any pending deletion indexes by retrieving active indexes from `/config/indexes` endpoint.


luchenyuxx (approved):

tests/integration/resources/cassettes/test_logs_indexes_order/TestLogsIndexesOrder.test_resource_import.yaml
```diff
@@ -8,7 +34,38 @@ interactions:
...
+    status:
+      code: 200
+      message: OK
+- request:
+    body: null
+    headers:
+      Content-Type:
+      - application/json
+    method: GET
+    uri: https://api.datadoghq.eu/api/v1/logs/config/indexes
```
skarimo, line 50: `test-index` is deleted from R1 and moved to tend of the list

HantingZhang2 (approved):

PR #250, opened 2024-05-22
Ensure clean exit on invalid keys

skarimo:


nkzou (approved):
`_handle_deprecated` still uses the renamed `_exit_cleanup` functions, I'm going to merge this PR regardless but you might want to make a followup to remove/fix the calls in the `_handle_deprecated`

PR #251, opened 2024-05-23
Add support for metric percentiles syncing

skarimo:
APITL-899


jhgilbert (approved):

datadog_sync/model/metric_percentiles.py
```diff
@@ -0,0 +1,58 @@
...
+    resource_type = "metric_percentiles"
+    resource_config = ResourceConfig(
+        base_path="/metric/distribution/summary_aggr",
+        excluded_attributes=["key"],
+    )
+    # Additional MetricPercentiles specific attributes
+    metrics_summaries_get_path = "/metric/distribution/list_summaries"
+
+    async def get_resources(self, client: CustomClient) -> List[Dict]:
+        resp = await client.get(self.metrics_summaries_get_path)
```
efraese, line 22: We should pass the `window` query param too to scope this query to just 14 days. It would look like this: `?window=86400`

efraese (dismissed):

PR #267, opened 2024-07-22
Handle ID clashes across resources

skarimo:
Previously, we assumed ID's across resources will be unique. However this is not true, rather ID's are unique within resources and can be duplicated in between. For example, `metric_tag_configurations` and `metric_metadata` both rely on the metric name.

This PR:

- updates internal handling to use tuple of (resource_type, _id) to build graph ensuring ids don't class
- Removes the previous resource manager which was unnecessary redirection and bit confusing


datadog_sync/utils/resources_handler.py
```diff
@@ -281,21 +286,70 @@ async def _pre_apply_hook_cb(self, resource_type: str) -> None:
...
+        missing_resources = set()
+
+        for resource_type, _id in self.config.state.get_all_resources(self.config.resources_arg).keys():
+            deps, missing = self._resource_connections(resource_type, _id)
+            dependency_graph[(resource_type, _id)] = deps
+            missing_resources.update(missing)
+
+        return dependency_graph, missing_resources
+
+    def _resource_connections(self, resource_type: str, _id: str) -> Tuple[Set[Tuple[str, str]], Set[Tuple[str, str]]]:
```
jack-edmonds-dd, line 315: We should probably document what this is returning. It says it's returning `failed_connections, missing_resources` but `failed_connections` is also treated as `deps` above.

jack-edmonds-dd (approved):

PR #271, opened 2024-09-04
HAMR-179 Add pyproject.toml to fix docker build

michael-richey:
### What does this PR do?
Fixes the `docker build` command which is currently broken.

### Description of the Change
Creates a skeleton `pyproject.toml` file so the build will work.

### Alternate Designs
We could use a different build backend. For example, passing `--use-pep517` would fix it but swapping the build backend seems like a larger change. Changing the base image (to python3.12-slim-bullseye) also seems to resolve but again it's a larger change. 

### Verification Process
Tested the documented `docker build` command then ran an `import` and `sync` with the newly created docker image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.11.1 / 2024-09-04
+
+### Fixed
+* fixed the `docker build` command
+### Changed
+* added the `pyproject.toml` file by @michael-richey in https://github.com/DataDog/datadog-sync-cli/pull/271
+
+**Full Changelog**: https://github.com/DataDog/datadog-sync-cli/compare/0.11.0...0.11.1
+
 ## 0.11.0 / 2024-06-27
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 ## 0.11.1 / 2024-09-04
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.11.1 / 2024-09-04
+
+### Fixed
+* fixed the `docker build` command
```
drichards-87, line 6:
```suggestion
* Fix the `docker build` command
```

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+### Changed
+* added the `pyproject.toml` file by @michael-richey in https://github.com/DataDog/datadog-sync-cli/pull/271
```
drichards-87, line 8:
```suggestion
* Add the `pyproject.toml` file by @michael-richey in https://github.com/DataDog/datadog-sync-cli/pull/271
```

drichards-87 (approved):
Left a couple of small suggestions from Docs and approved the PR.

CHANGELOG.md
michael-richey: Changed to match yours

CHANGELOG.md
michael-richey: Changed to match yours

heyronhay (dismissed):

skarimo (approved):

PR #283, opened 2024-09-12
Adding observability metrics to sync-cli

michael-richey:
### What does this PR do?
Add metrics to sync-cli so that we can build dashboards for users to see the last time resources were synchronized. Add a boolean flag to the command line `--send-metrics` that defaults to `true` but that users can set to false if they don't want these metrics. Also fixed a bug with the notebooks resource batch size by switching it from 500 to 100.

### Description of the Change
This is a best effort attempt to send metrics for a handful of commands to both the source and the destination. If we can send the information to both source and destination then you can use a dashboard on either org to see the sync metrics.

### Possible Drawbacks

Most users won't be passing the destination url, api key, and app key to the import command or passing the source url, api key, and app key to the sync command. So it's likely that the metrics for imports will only be at the source and the metrics for syncs will only be at the destination.


.gitignore
```diff
@@ -1,6 +1,7 @@
 datadog_sync/version.py
 **/*.pyc
 **/*.pyo
+**/*.swp
```
michael-richey, line 4: swap files created by `vi`

datadog_sync/commands/_import.py
```diff
@@ -6,12 +6,13 @@
...
 
 from datadog_sync.constants import Command
-from datadog_sync.commands.shared.options import common_options, source_auth_options
+from datadog_sync.commands.shared.options import common_options, destination_auth_options, source_auth_options
 from datadog_sync.commands.shared.utils import run_cmd
 
 
 @command(Command.IMPORT.value, short_help="Import Datadog resources.")
 @source_auth_options
+@destination_auth_options
```
michael-richey, line 15: Allows us to pass in destination url, api key, and app key to `import` command. We can use these to send metrics to source and destination. These aren't required so you can still run an `import` without them.

datadog_sync/commands/shared/options.py
```diff
@@ -164,6 +164,14 @@ def click_config_file_provider(ctx: Context, opts: CustomOptionClass, value: Non
...
     ),
+    option(
+        "--send-metrics",
+        type=bool,
+        required=False,
+        default=True,
+        help="Enables sync-cli metrics being sent to both source and destination",
+        cls=CustomOptionClass,
+    ),
 ]
```
michael-richey, line 175: Allow the user to turn off these metrics

datadog_sync/model/notebooks.py
```diff
@@ -27,7 +27,7 @@ class Notebooks(BaseResource):
     )
     # Additional Notebooks specific attributes
     pagination_config = PaginationConfig(
-        page_size=500,
+        page_size=100,
```
michael-richey, line 30: 500 times out, 100 works

tests/integration/helpers.py
```diff
@@ -36,11 +36,17 @@ def test_resource_import(self, runner, caplog):
...
 
         ret = runner.invoke(
-            cli, ["import", "--validate=false", f"--resources={self.resource_type}", f"--filter={self.filter}"]
+            cli,
+            [
+                "import",
+                "--validate=false",
+                f"--resources={self.resource_type}",
+                f"--filter={self.filter}",
+            ],
```
michael-richey, line 45: These are the results of running `black` on this file

PR #287, opened 2024-09-27
HAMR-164: Add migrate command

michael-richey:
### What does this PR do?
Add a `migrate` command that both imports and sync resources in one step.

### Description of the Change
Add the `migrate` command that just does an `import` and a `sync`. Renamed the non_import_common_options to be diff_common_options since these options are shared between diff, sync, and migrate. Created sync_common_options to be shared between sync and migrate.


datadog_sync/commands/shared/options.py
```diff
@@ -94,7 +94,7 @@ def click_config_file_provider(ctx: Context, opts: CustomOptionClass, value: Non
         type=int,
         default=60,
         show_default=True,
-        help="The HTTP request retry timeout period. Defaults to 60s",
+        help="The HTTP request retry timeout period in seconds.",
```
michael-richey, line 97: the `show_default=True` already shows the default of 60.

datadog_sync/commands/shared/options.py
```diff
@@ -104,7 +104,7 @@ def click_config_file_provider(ctx: Context, opts: CustomOptionClass, value: Non
         type=int,
         default=30,
         show_default=True,
-        help="The HTTP request timeout period. Defaults to 30s",
+        help="The HTTP request timeout period in seconds.",
```
michael-richey, line 107: the `show_default=True` already shows the default of 30.

datadog_sync/commands/shared/options.py
```diff
@@ -199,6 +203,29 @@ def click_config_file_provider(ctx: Context, opts: CustomOptionClass, value: Non
...
+        default=False,
+        show_default=True,
+        help="Force importing and syncing resources that could be potential dependencies to the requested resources.",
+        cls=CustomOptionClass,
+    ),
+    option(
+        "--create-global-downtime",
+        required=False,
+        is_flag=True,
+        default=False,
```
heyronhay, line 220: we should probably switch the create-global-downtime to default to true since we almost always want it to happen.

heyronhay (approved):

PR #290, opened 2024-10-04
Change the default value of --create-global-downtime

michael-richey:
### What does this PR do?
Create a global downtime at the destination for `sync` and `migrate` commands by default. Allow the user to override it by setting the `--create-global-downtime` command line option to `False`

### Description of the Change

This changes the default value of `--create-global-downtime` to `True`, changed the type to `bool` rather than a flag.

### Possible Drawbacks

This changes the default behavior on `sync-cli`. Users could accidentally create downtimes at destinations when they don't mean to do so. This will mute *all their monitors* at the destination org.

### Release Notes

Need to bump the major release number for this change.

### DIFF
--- a/datadog_sync/commands/shared/options.py
+++ b/datadog_sync/commands/shared/options.py
@@ -227,8 +227,8 @@ _sync_common_options = [
     option(
         "--create-global-downtime",
         required=False,
-        is_flag=True,
-        default=False,
+        type=bool,
+        default=True,
         show_default=True,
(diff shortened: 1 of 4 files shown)


heyronhay (approved):
Looks like it already has logic to prevent multiple global downtimes from being created, is that correct?

michael-richey:
> Looks like it already has logic to prevent multiple global downtimes from being created, is that correct?

Exactly!

PR #292, opened 2024-10-09
Allow the resources directories to be passed in

michael-richey:
### What does this PR do?
Allows users to specify the directories for storing resources locally.

### Description of the Change
Added `--source-resources-path` and `--destination-resources-path` flags to the command line rather than hard coding the local file storage. The defaults remain the same.

### Verification Process
Added tests and ran tests locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ By default all commands check the Datadog Disaster Recovery (DDR) status of both
 
 #### State files
 
-By default a `resources` directory is generated in the current working directory of the user. This directory contains `json` mapping of resources between the source and destination organization. To avoid duplication and loss of mapping, this directory should be retained between tool usage. To override these directories use the `--source-resources-path` and `--destination-resource-path`.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -208,7 +208,7 @@ By default all commands check the Datadog Disaster Recovery (DDR) status of both
 
 #### State files
 
-A `resources` directory is generated in the current working directory of the user. This directory contains `json` mapping of resources between the source and destination organization. To avoid duplication and loss of mapping, this directory should be retained between tool usage.
+By default a `resources` directory is generated in the current working directory of the user. This directory contains `json` mapping of resources between the source and destination organization. To avoid duplication and loss of mapping, this directory should be retained between tool usage. To override these directories use the `--source-resources-path` and `--destination-resource-path`.
```
estherk15, line 211:
```suggestion
By default, a `resources` directory is generated in the current working directory of the user. This directory contains `json` mapping of resources between the source and destination organization. To avoid duplication and loss of mapping, this directory should be retained between tool usage. To override these directories use the `--source-resources-path` and `--destination-resource-path`.
```

estherk15 (dismissed):

heyronhay (approved):
Good idea!

PR #310, opened 2025-01-06
Add support for writing to an S3 bucket

michael-richey:
### What does this PR do?

Add support for writing to an S3 bucket via command line arguments or env vars.

### Description of the Change

Created a new extension to the base state for AWS S3.


heyronhay (approved):

datadog_sync/commands/shared/options.py
```diff
@@ -209,10 +209,72 @@ def click_config_file_provider(ctx: Context, opts: CustomOptionClass, value: Non
...
     ),
+    option(
+        "--use-aws-s3-bucket",
+        required=False,
+        type=bool,
+        default=False,
+        show_default=True,
+        help="Enables using an AWS S3 bucket to store resources' state",
+        cls=CustomOptionClass,
+    ),
```
skarimo, line 224: Nit: It would be nicer to make this `click.Choice()` with `[file, aws_s3_bucket]` rather than a bucket in case we expand the supported storage in the future

michael-richey: I see you used it for the `--cleanup` option, I'll use that as an example. Thank you!

michael-richey: @skarimo I changed to use `click.Choice()` You were right, I think that made it a lot nicer.

Issue #324, opened 2025-02-19
Why is --verify-ddr-status=true the default for the import command?


shuheiktgw:
To me, it’s not very intuitive why `--verify-ddr-status=true` is the default for the import command, since, as I understand it, that requires configuring destinations. I feel it should be `false` by default for the import command. What do you think? If we agree, I’d be happy to change the default to false.

shuheiktgw:
If we don't we still need to update [this example](https://github.com/DataDog/datadog-sync-cli/blob/df1a92a1a9aae8bc71d47388c7d5d5534b11a4f5/README.md?plain=1#L42-L46).

michael-richey:
If there is a disaster and DDR is activated or unavailable then we should not sync those resources because they're "suspect" at that point. For example during a disaster you might copy over a corrupted dashboard over your good working dashboard.
However, I agree that from the import command users should not have to set up their destinations (and conversely you should not have to setup a source to run the sync command).

PR #335, opened 2025-03-20
Narrow the scope of DDR checking based on the command

michael-richey:
### What does this PR do?

Addressing a component of [issue 324](https://github.com/DataDog/datadog-sync-cli/issues/324). DDR checking now checks source or destination based on the command.

### Description of the Change

If the user is running the `sync` command we should only check the DDR status of the destination.
If the user is running the `import` command we should only check the DDR status of the destination.

### Alternate Designs

Defaulting DDR to `false` - This idea is rejected because running `import` from a source (or `sync` to a destination) that is having a disaster could lead to corrupted or bad data. It's import to check for a disaster first, during a disaster sync-cli may not be to connect at all, so over riding this dangerous condition needs to be the default.

### Possible Drawbacks

Users that are only using sync-cli for a single migration or using it as a backup will have have to pass the extra flag.


shuheiktgw (approved):
This looks great to me, thanks! 🙂

heyronhay (approved):

PR #337, opened 2025-04-01
Fix to respect user provided s3 source and destination path prefixes

sahu-sunil:
### What does this PR do?

User provided aws-bucket-key-prefix-source and aws-bucket-key-prefix-destination were not respected instead default paths were taken. This MR fixes the issue.
### Description of the Change

While creating the `State` object we have evaluated the correct source and destination paths [here](https://github.com/DataDog/datadog-sync-cli/blob/main/datadog_sync/utils/configuration.py#L183), and trying to [get from their envvar](https://github.com/DataDog/datadog-sync-cli/blob/main/datadog_sync/utils/state.py#L32) names in `State` which looks incorrect

### Review checklist (to be filled by reviewers)

- [ ] Feature or bug fix MUST have appropriate tests (unit, integration, etc...)
- [ ] PR title must be written as a CHANGELOG entry [(see why)](/CONTRIBUTING.md#pull-request-title)
- [ ] Files changes must correspond to the primary purpose of the PR as described in the title (small unrelated changes should have their own PR)
- [ ] PR must have one `changelog/` label attached. If applicable it should have the `backward-incompatible` label attached.
- [ ] PR should not have `do-not-merge/` label attached.
- [ ] If Applicable, issue must have `kind/` and `severity/` labels attached at least.


sahu-sunil:
@michael-richey Could you please review this?

michael-richey:
> User provided  aws-bucket-key-prefix-source and aws-bucket-key-prefix-destination were not respected

Can you explain this a little further? I've run things like the following without issue:
```
datadog-sync migrate --resources="dashboards" --storage-type=s3 --aws-bucket-name=test --aws-bucket-key-prefix-source=myprefix/source  --aws-bucket-key-prefix-destination=myprefix/destination --aws-region-name=us-east-1 
```

If I can replicate what you're seeing then we might be able to resolve the issue rather than eliminating the use of the env vars which is what it appears this PR does. Eliminating the env vars might still be the solution, but I'd just like to understand the problem you're facing better.

PR #533, opened 2026-04-20
feat: migrate Tier 2 resources and fix 409 conflicts

michael-richey:
## Summary
- Migrate Tier 2 resources (`roles`, `security_monitoring_rules`, `team_memberships`) to `map_existing_resources`
- Fix 409 conflict handling for `logs_metrics` and `synthetics_global_variables`

These changes were originally in PRs #529 and #530 which were stacked on #528. PR #528 was squash-merged to main, but the stacked PRs only merged into their intermediate branches. This PR cherry-picks those missing commits onto main.

## Test plan
- [x] CI passes
- [x] Verify Tier 2 resources use `map_existing_resources` correctly
- [x] Verify 409 conflicts are handled for `logs_metrics`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #537, opened 2026-04-20
fix: exclude reserved Datadog built-in roles from all sync operations

michael-richey:
## Summary
- Adds `BUILTIN_ROLE_NAMES` constant (`Datadog Admin Role`, `Datadog Read Only Role`, `Datadog Standard Role`) to `roles.py`
- Overrides `filter()` to skip these roles at import and sync time
- Guards `delete_resource()` to raise `SkipResource` at cleanup time

## Problem
The three built-in Datadog roles are reserved and managed by Datadog — the API returns 400 when trying to create/update them and will refuse deletion. Previously sync-cli would fail with 400/403 errors when these roles appeared in source state.

## Test plan
- [x] `import` no longer pulls in built-in roles
- [x] `sync` skips built-in roles from source state
- [x] `sync --cleanup=force` does not attempt to delete built-in roles from destination

### DIFF
--- a/datadog_sync/model/roles.py
+++ b/datadog_sync/model/roles.py
@@ -16,6 +16,17 @@ if TYPE_CHECKING:
     from datadog_sync.utils.custom_client import CustomClient
 
 
+# These role names are reserved and managed by Datadog — they cannot be created,
+# updated, or deleted via the API and must be excluded from all sync operations.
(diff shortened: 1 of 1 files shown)


michael-richey:
Superseded by commit 25a1592b on PR #533, which applies this fix onto the tier 2 migrations branch.

PR #539, opened 2026-04-21
fix: skip custom check monitors missing options.groupby at sync time

michael-richey:
## Summary

- Custom check (service check) monitors were failing sync with `400: "A grouping must be specified for custom checks"`
- Root cause: source monitors have an empty or missing `options.groupby`; the destination API requires at least one entry for this monitor type
- Fix: detect the condition in `pre_resource_action_hook` on `Monitors`, skip at sync time with actionable message

## Changes

- `datadog_sync/model/monitors.py`: add `pre_resource_action_hook` that checks service check monitors for a non-empty `options.groupby`
- `tests/unit/test_monitors.py`: 7 unit tests covering all field/type/null edge cases

## Test plan

- [x] `pytest tests/unit/` — all unit tests pass

### DIFF
--- a/datadog_sync/model/monitors.py
+++ b/datadog_sync/model/monitors.py
@@ -79,7 +79,7 @@ class Monitors(BaseResource):
                 raise SkipResource(
                     _id,
                     self.resource_type,
-                    "Custom check monitor requires at least one group-by. "
+                    "Deprecated resource configuration: Custom check monitor requires at least one group-by. "
                     "Update the source monitor's options.groupby before syncing.",
                 )
 
--- a/datadog_sync/model/monitors.py
+++ b/datadog_sync/model/monitors.py
@@ -73,7 +73,15 @@ class Monitors(BaseResource):
         return str(resource["id"]), resource
 
     async def pre_resource_action_hook(self, _id, resource: Dict) -> None:
-        pass
+        if resource.get("type") == "service check":
+            groupby = resource.get("options", {}).get("groupby", [])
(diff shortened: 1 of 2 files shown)


datadog_sync/model/monitors.py
```diff
@@ -73,7 +73,15 @@ async def import_resource(self, _id: Optional[str] = None, resource: Optional[Di
...
 
     async def pre_resource_action_hook(self, _id, resource: Dict) -> None:
-        pass
+        if resource.get("type") == "service check":
+            groupby = (resource.get("options") or {}).get("groupby", [])
+            if not groupby:
+                raise SkipResource(
+                    _id,
+                    self.resource_type,
+                    "Custom check monitor requires at least one group-by. "
```
heyronhay: I wonder if we should emit a common message for this kind of thing, like "DEPRECATED RESOURCE SETTING" or something?

heyronhay (dismissed):

datadog_sync/model/monitors.py
michael-richey: Yeah, I was even thinking a totally different class. Not Successful, Skipped, or Failed. Skipping is more for resources that don't need updating. It's actually closer to Failed, but I like the idea of Deprecated.

heyronhay: or "Needs Update"? Something along those lines. Makes sense to put it in a different category. "Cannot Sync"?

heyronhay (approved):

PR #544, opened 2026-04-23
feat(sync): add --minimize-reads with type-scoped storage loading

michael-richey:
## Summary

Stacked on #543.

For large orgs (10,000+ resources), sync-cli loads all resource files from cloud storage even when syncing a small subset (e.g. `--resources=roles`). This causes managed-sync Phase 2/3 to take ~25 minutes each, leaving no time for Phase 6.

Adds `--minimize-reads` flag (sync command only) that scopes storage reads to only the requested resource type(s):

- **Type-scoped strategy**: when `--resources=roles`, only list/fetch `roles.*` files instead of all 10,000 files
- For `--resources=roles` with 10,000 total resources: reduces reads from ~20,000 to ~100

**Constraints** (enforced at CLI parse time):
- Requires `--resource-per-file`
- Requires `--resources`
- Restricted to `sync` command only
- Must not be combined with `--cleanup`

ID-targeted loading (exact filter matches → direct key fetch, no listing) added in the next PR.

## Test plan
- [x] `pytest tests/unit/test_minimize_reads_type_scoped.py` — 10 tests pass
- [x] `pytest tests/unit/` — full regression, 360 tests pass
- [x] Verify `diffs`/`import` commands reject `--minimize-reads` with "no such option"

### DIFF
--- a/datadog_sync/utils/configuration.py
+++ b/datadog_sync/utils/configuration.py
@@ -314,8 +314,10 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
             raise click.UsageError("--minimize-reads requires --resource-per-file")
         if not kwargs.get("resources", None):
             raise click.UsageError("--minimize-reads requires --resources")
+        if kwargs.get("cleanup") and kwargs["cleanup"].lower() in ("true", "force"):
+            raise click.UsageError("--minimize-reads cannot be combined with --cleanup")
 
-    # Determine type-scoped loading (ID-targeted loading added in PR 3)
+    # Determine loading strategy for minimize-reads
     _state_resource_types = None  # None = full load (existing behavior)
     if minimize_reads and (rs := kwargs.get("resources", None)):
         _state_resource_types = [r.strip().lower() for r in rs.split(",") if r.strip()]
--- a/datadog_sync/utils/configuration.py
+++ b/datadog_sync/utils/configuration.py
@@ -306,6 +306,20 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
         )
 
     resource_per_file = kwargs.get(RESOURCE_PER_FILE, False)
+    minimize_reads = kwargs.get("minimize_reads", False)
+
+    # Validate --minimize-reads constraints
+    if minimize_reads:
+        if not resource_per_file:
+            raise click.UsageError("--minimize-reads requires --resource-per-file")
+        if not kwargs.get("resources", None):
+            raise click.UsageError("--minimize-reads requires --resources")
+
+    # Determine type-scoped loading (ID-targeted loading added in PR 3)
+    _state_resource_types = None  # None = full load (existing behavior)
+    if minimize_reads and (rs := kwargs.get("resources", None)):
+        _state_resource_types = [r.strip().lower() for r in rs.split(",") if r.strip()]
+
     # Initialize state
     state = State(
         type_=storage_type,
@@ -313,7 +327,11 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
         destination_resources_path=destination_resources_path,
         config=config,
         resource_per_file=resource_per_file,
+        resource_types=_state_resource_types,  # None = full load
     )
+
+    if _state_resource_types is not None:
+        logger.info(f"minimize-reads: type-scoped loading for {_state_resource_types}")
 
     # Initialize Configuration
     config = Configuration(
--- a/datadog_sync/utils/state.py
+++ b/datadog_sync/utils/state.py
@@ -22,6 +22,8 @@ from datadog_sync.utils.storage.storage_types import StorageType
 
 class State:
     def __init__(self, type_: StorageType = StorageType.LOCAL_FILE, **kwargs: object) -> None:
+        self._resource_types = kwargs.get("resource_types", None)
+        self._minimize_reads = self._resource_types is not None
         resource_per_file = kwargs.get(RESOURCE_PER_FILE, False)
         source_resources_path = kwargs.get(SOURCE_PATH_PARAM, SOURCE_PATH_DEFAULT)
         destination_resources_path = kwargs.get(DESTINATION_PATH_PARAM, DESTINATION_PATH_DEFAULT)
@@ -76,7 +78,8 @@ class State:
         return self._data.destination
 
     def load_state(self, origin: Origin = Origin.ALL) -> None:
-        self._data = self._storage.get(origin)
+        # MODIFIED: resource_types param added. None = full load (existing behavior).
+        self._data = self._storage.get(origin, resource_types=self._resource_types)
 
     def dump_state(self, origin: Origin = Origin.ALL) -> None:
         self._storage.put(origin, self._data)
--- a/datadog_sync/utils/state.py
+++ b/datadog_sync/utils/state.py
@@ -78,7 +78,7 @@ class State:
         return self._data.destination
 
     def load_state(self, origin: Origin = Origin.ALL) -> None:
-        # MODIFIED: resource_types param added. None = full load (existing behavior).
+        # resource_types=None → load all types (default behavior)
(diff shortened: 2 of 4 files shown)


michael-richey (commented):
Staff review of `--minimize-reads` (type-scoped strategy). The design is clean: scoping storage reads to requested resource types is the right approach, and the fallback semantics (None = full load) preserve backward compatibility perfectly. The `_check_id_collisions` fix from log-only to active skip is a correct behavior change. Test coverage is solid for the happy path.

**Critical**
- `--minimize-reads` + `--cleanup` combination is undocumented in code and unenforceable by help text alone. The code only validates `resource_per_file` and `resources`; the cleanup incompatibility is absent. This is a data-loss risk: cleanup compares ALL destination resources against scoped source state and would delete everything not in scope.

**Significant**
- "MODIFIED" comment in `state.py`'s `load_state` is a dev note, not a permanent code comment.
- "ID-targeted loading added in PR 3" in `configuration.py` leaks PR numbering into production code.

**Minor**
- `LocalFile.get()` with `resource_types` scoping still scans the full directory (O(N) `os.listdir`), just skipping file opens. The performance win is real but "reduces reads" in the PR description overstates the LocalFile gain — for S3/GCS/Azure it's a genuine listing reduction, for local it's a file-open reduction.
- No test for `--minimize-reads --cleanup` being rejected (since it isn't).

Inline comments on specific lines below.

datadog_sync/utils/configuration.py
```diff
@@ -306,15 +306,33 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
...
 
     resource_per_file = kwargs.get(RESOURCE_PER_FILE, False)
+    minimize_reads = kwargs.get("minimize_reads", False)
+
+    # Validate --minimize-reads constraints
+    if minimize_reads:
+        if not resource_per_file:
+            raise click.UsageError("--minimize-reads requires --resource-per-file")
+        if not kwargs.get("resources", None):
+            raise click.UsageError("--minimize-reads requires --resources")
```
michael-richey, line 316: **Missing validation: `--minimize-reads` + `--cleanup` is a data-loss risk.**

The help text says 'Must not be combined with --cleanup' but there is no enforcement here. When `--cleanup` is active, `get_resources_to_cleanup` in State iterates ALL destination resources and compares them against source. With type-scoped loading, source only contains the requested type(s) — every destination resource of other types appears 'missing from source' and gets queued for deletion.

Add the check here alongside the other validations:

```python
if kwargs.get('cleanup') and kwargs['cleanup'].lower() in ('true', 'force'):
    raise click.UsageError('--minimize-reads cannot be combined with --cleanup')
```

And add a corresponding test in `TestMinimizeReadsCLIValidation`.

datadog_sync/utils/state.py
```diff
@@ -76,7 +78,8 @@ def destination(self):
         return self._data.destination
 
     def load_state(self, origin: Origin = Origin.ALL) -> None:
-        self._data = self._storage.get(origin)
+        # MODIFIED: resource_types param added. None = full load (existing behavior).
```
michael-richey: This comment reads like a PR development note ('MODIFIED: resource_types param added') rather than permanent documentation. Since the function signature change is visible in the code itself, the comment adds noise without value. Either remove it or rephrase as a contract note:

```python
# resource_types=None → load all types (default behavior)
```

datadog_sync/utils/configuration.py
```diff
@@ -306,15 +306,33 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
...
+
+    # Determine type-scoped loading (ID-targeted loading added in PR 3)
```
michael-richey: 'ID-targeted loading added in PR 3' is a PR reference, not a code comment. Once this merges, 'PR 3' is meaningless to future readers. Replace with a description of the code's intent:

```python
# Determine loading strategy. Default is full load (resource_types=None).
# With --minimize-reads, fall back to type-scoped listing.
# ID-targeted loading (exact filter → direct key fetch) is handled in build_config.
```

tests/unit/test_minimize_reads_type_scoped.py
```diff
@@ -0,0 +1,202 @@
...
+    def test_minimize_reads_not_available_on_import_command(self, runner):
+        """--minimize-reads must not be accepted by the import command."""
+        result = runner.invoke(
+            cli,
+            ["import", "--minimize-reads", "--source-api-key=x"],
+        )
+        assert result.exit_code != 0
+        assert "no such option" in result.output.lower()
+
+
```
michael-richey, line 163: The help text documents 'Must not be combined with --cleanup' but the tests don't cover that constraint — and as noted in the inline comment on `configuration.py`, the validation isn't actually enforced in code yet. Add a test here:

```python
def test_minimize_reads_cannot_be_combined_with_cleanup(self, runner):
    result = runner.invoke(
        cli,
        [
            'sync',
            '--minimize-reads',
            '--resource-per-file',
            '--resources=roles',
            '--cleanup=Force',
            '--source-api-key=x',
            '--destination-api-key=y',
        ],
    )
    assert result.exit_code != 0
    assert 'cleanup' in (result.output + str(result.exception)).lower()
```

This test will fail until the enforcement is added, making the missing guard visible in CI.

datadog_sync/utils/storage/_base_storage.py
```diff
@@ -40,27 +40,32 @@ def _sanitize_id_for_filename(resource_id: str) -> str:
         return resource_id.replace(":", ".")
 
     @staticmethod
-    def _check_id_collisions(resource_data: dict, resource_type: str) -> None:
-        """Log an error for any two IDs that sanitize to the same filename segment.
+    def _check_id_collisions(resource_data: dict, resource_type: str) -> set:
```
michael-richey, line 43: The 'first ID encountered wins' behavior now relies on Python dict insertion-order guarantee (Python 3.7+). This is fine for the supported Python versions, but worth making it explicit in the docstring since it determines which resource's data survives a collision:

```python
# Dict iteration is insertion-ordered (Python 3.7+). The first ID
# encountered when iterating resource_data is the one that wins;
# all subsequent IDs that map to the same sanitized filename are skipped.
```

Also: when the same collision exists in both source and destination, this method is called twice and logs the error twice. Consider whether deduplication is worth it (probably not for a rare edge case, but worth acknowledging).

tests/unit/test_minimize_reads_type_scoped.py
michael-richey: Fixed. Added `test_minimize_reads_cannot_be_combined_with_cleanup` to `TestMinimizeReadsCLIValidation`.

datadog_sync/utils/configuration.py
michael-richey: Fixed. Added the  guard in `configuration.py` alongside the existing `--resources` validation check. Also added `test_minimize_reads_cannot_be_combined_with_cleanup` to `TestMinimizeReadsCLIValidation`.

datadog_sync/utils/configuration.py
michael-richey: Fixed. Replaced with `# Determine loading strategy for minimize-reads`.

datadog_sync/utils/state.py
michael-richey: Fixed. Updated to `# resource_types=None → load all types (default behavior)`.

datadog_sync/utils/storage/_base_storage.py
michael-richey: Fixed on `hamr/minimize-reads-1-sanitize`. Added the insertion-order comment above the `seen` dict in `_check_id_collisions`: 'Dict iteration is insertion-ordered (Python 3.7+). The first ID encountered wins; subsequent colliders are skipped (only logged).'

riyazsh (dismissed):

riyazsh (approved):

PR #545, opened 2026-04-23
feat(sync): add ID-targeted loading and per-dependency lazy loading

michael-richey:
## Summary

Stacked on #544.

Extends `--minimize-reads` with a faster strategy and lazy dependency loading:

**ID-targeted loading (fast path):** When all filters use `Name=id+ExactMatch+OR`, constructs storage keys directly and fetches only those files — no listing needed. For managed-sync's typical invocation (100 resources with exact ID filters), reduces reads from ~20,000 to ~204 (100 files + up to 100 dependency files, source+destination).

**Lazy dependency loading:** `State.ensure_resource_loaded(type, id)` lazily fetches cross-type dependencies when encountered in `_resource_connections()`. Loads both source and destination state so `connect_id()` can remap IDs correctly.

**Strategy selection** (automatic, in `build_config()`):
1. ID-targeted: all filters use `Name=id+ExactMatch+OR` → direct key fetch
2. Type-scoped fallback: non-ID filters or no filters → type-prefix listing (PR 2)

**Note:** `get_by_ids()` requires `--resource-per-file` and raises `ValueError` if called without it. All `State` instantiations that pass `exact_ids` must also pass `resource_per_file=True`.

**Performance (10k dashboards + 10k monitors, 1-to-1 deps):**
| Command | Before | After |
|---------|--------|-------|
| `--resources=dashboards --filter=...Name=id...` (100 IDs) | 40,000 reads | ~204 reads |
| `--resources=roles` (500 roles, type-scoped) | 40,000 reads | ~1,000 reads |

**Known limitation:** `monitors.py` and `restriction_policies.py` have `connect_id()` overrides that access hardcoded destination types beyond `resource_to_connect`. With `--minimize-reads`, those types return empty dicts (no crash, just silent ID-remapping skip). Testing should confirm the specific types managed-sync syncs don't hit this.

## Test plan
- [x] `pytest tests/unit/test_minimize_reads_id_targeted.py` — 19 tests pass
- [x] `pytest tests/unit/` — full regression, 379 tests pass

### DIFF
--- a/datadog_sync/utils/configuration.py
+++ b/datadog_sync/utils/configuration.py
@@ -363,7 +363,7 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
 
     # Determine loading strategy for minimize-reads
     _state_resource_types = None  # type-scoped; None = full load (existing behavior)
-    _state_exact_ids = None       # ID-targeted; None = not using ID-targeted
+    _state_exact_ids = None  # ID-targeted; None = not using ID-targeted
     if minimize_reads and (rs := kwargs.get("resources", None)):
         raw_types = [r.strip().lower() for r in rs.split(",") if r.strip()]
         # Try ID-targeted strategy first (fast path: exact IDs from filters)
@@ -382,7 +382,7 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
         config=config,
         resource_per_file=resource_per_file,
         resource_types=_state_resource_types,  # None = full load or ID-targeted
-        exact_ids=_state_exact_ids,             # None = not using ID-targeted
+        exact_ids=_state_exact_ids,  # None = not using ID-targeted
     )
 
     if _state_exact_ids is not None:
--- a/datadog_sync/utils/configuration.py
+++ b/datadog_sync/utils/configuration.py
@@ -138,8 +138,8 @@ class Configuration(object):
 def _unwrap_exact_match_pattern(pattern: str) -> str:
     """Extract the raw ID value from an ExactMatch ^...$-wrapped regex pattern.
 
-    Raises ValueError if the pattern is not ^...$-wrapped, so callers can
-    gracefully fall back to type-scoped loading.
+    Defensive check: ExactMatch always produces ^...$, so ValueError should not
+    fire in practice. Raises ValueError so callers can detect unexpected patterns.
     """
     if not (pattern.startswith("^") and pattern.endswith("$")):
         raise ValueError(f"Expected ExactMatch regex ^...$, got: {pattern!r}")
@@ -170,7 +170,9 @@ def extract_exact_id_filters(
         # All filters must be id-field ExactMatch
         if not all(f.attr_name == ["id"] and f.operator == EXACT_MATCH_OPERATOR for f in rt_filters):
             return None
-        # Extract raw IDs from ^...$-wrapped regex patterns
+        # Extract raw IDs from ^...$-wrapped regex patterns.
+        # Defensive: ExactMatch guarantees ^...$, so ValueError should not fire.
+        # Kept as a safety net in case filter construction changes upstream.
         try:
             ids = [_unwrap_exact_match_pattern(f.attr_re.pattern) for f in rt_filters]
         except ValueError:
@@ -372,6 +374,7 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
         _state_exact_ids = extract_exact_id_filters(early_filters, filter_operator, raw_types)
         if _state_exact_ids is None:
             # Fall back to type-scoped loading
+            logger.debug("minimize-reads: ID-targeted not eligible — filters are not all id+ExactMatch+OR")
             _state_resource_types = raw_types
 
     # Initialize state
--- a/datadog_sync/utils/configuration.py
+++ b/datadog_sync/utils/configuration.py
@@ -41,7 +41,7 @@ from datadog_sync.model.downtime_schedules import DowntimeSchedules
 from datadog_sync.utils.custom_client import CustomClient
 from datadog_sync.utils.base_resource import BaseResource
 from datadog_sync.utils.log import Log
-from datadog_sync.utils.filter import Filter, process_filters
+from datadog_sync.utils.filter import Filter, process_filters, EXACT_MATCH_OPERATOR
 from datadog_sync.utils.resource_utils import CustomClientHTTPError
 from datadog_sync.utils.state import State
 from datadog_sync.utils.storage.storage_types import StorageType
@@ -133,6 +133,50 @@ class Configuration(object):
     async def exit_async(self):
         await self.source_client._end_session()
         await self.destination_client._end_session()
+
+
+def _unwrap_exact_match_pattern(pattern: str) -> str:
+    """Extract the raw ID value from an ExactMatch ^...$-wrapped regex pattern.
+
+    Raises ValueError if the pattern is not ^...$-wrapped, so callers can
+    gracefully fall back to type-scoped loading.
+    """
+    if not (pattern.startswith("^") and pattern.endswith("$")):
+        raise ValueError(f"Expected ExactMatch regex ^...$, got: {pattern!r}")
+    return pattern[1:-1]
+
+
+def extract_exact_id_filters(
+    filters: Dict[str, list],
+    filter_operator: str,
+    resource_types: list,
+) -> Optional[Dict[str, list]]:
+    """Return {type: [id1, id2, ...]} when all conditions allow ID-targeted loading.
+
+    Conditions (all must be true):
+    - filter_operator is OR (case-insensitive)
+    - Every resource type in resource_types has at least one filter
+    - All filters for each type use Name=id + ExactMatch operator
+
+    Returns None if any condition fails → caller falls back to type-scoped loading.
+    """
+    if filter_operator.lower() != "or":
+        return None
+    result = {}
+    for rt in resource_types:
+        rt_filters = filters.get(rt, [])
+        if not rt_filters:
+            return None  # No filters for this type — can't use ID-targeted
+        # All filters must be id-field ExactMatch
+        if not all(f.attr_name == ["id"] and f.operator == EXACT_MATCH_OPERATOR for f in rt_filters):
+            return None
+        # Extract raw IDs from ^...$-wrapped regex patterns
+        try:
+            ids = [_unwrap_exact_match_pattern(f.attr_re.pattern) for f in rt_filters]
+        except ValueError:
+            return None  # Pattern wasn't ^...$-wrapped — fall back gracefully
+        result[rt] = ids
+    return result
 
 
 def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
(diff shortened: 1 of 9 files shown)


datadog_sync/utils/state.py
```diff
@@ -78,8 +83,38 @@ def destination(self):
...
+        Contract:
+        - Idempotent: no-op if resource_id already in source state
+        - No-op when not in minimize-reads mode (_minimize_reads=False)
+        - Appends to state: never replaces existing entries
+        - Missing file: (None, None) → resource stays absent (correct behavior)
+        - asyncio-safe: fully synchronous, no await points
+        """
+        if not self._minimize_reads:
+            return
+        if resource_id in self._data.source[resource_type]:
```
michael-richey: **Performance bug: repeated file-not-found lookups when a dependency is absent.**

The idempotency guard checks `source`, so when a dependency file doesn't exist (`get_single` returns `(None, None)`), the resource is never inserted into `source`. Every subsequent call to `ensure_resource_loaded` for that same ID skips the early-return and issues another `get_single`. If a dashboard references a missing role and there are 100 dashboards, that's 100 unnecessary file-stat/S3-HEAD operations for one absent role.

Fix: track attempted IDs in a separate set regardless of whether they were found:

```python
# in __init__:
self._ensure_attempted: set[tuple[str, str]] = set()

def ensure_resource_loaded(self, resource_type: str, resource_id: str) -> None:
    if not self._minimize_reads:
        return
    key = (resource_type, resource_id)
    if key in self._ensure_attempted:
        return
    self._ensure_attempted.add(key)
    # existing get_single logic...
```

michael-richey (commented):
Staff review of the ID-targeted loading + lazy dependency loading feature. The overall design is clean and the performance wins (40k → ~204 reads for the managed-sync use case) are compelling. Strategy selection in `build_config` is well-structured and the fallback chain is easy to follow. A few issues worth addressing before merge.

**Critical**
- `ensure_resource_loaded` has a repeated-miss performance problem when a dependency file doesn't exist at all — `get_single` is called on every invocation for that ID forever.

**Significant**
- `get_by_ids` is copy-pasted identically across S3, GCS, and Azure. It belongs in the base class.
- `get_single` ignores the `origin` parameter and always reads both source and destination, wasting I/O when only one side is needed.
- The `ValueError` catch in `extract_exact_id_filters` is unreachable given the surrounding `f.operator == EXACT_MATCH_OPERATOR` guard.

**Minor**
- No debug/info log when ID-targeted loading falls back to type-scoped — makes troubleshooting harder.
- Known `connect_id()` override limitation (monitors, restriction_policies) should surface a `log.warning` rather than silently skip, since silent ID-remapping failure is hard to detect in production.

Inline comments on the specific locations below.

datadog_sync/utils/state.py
```diff
@@ -78,8 +83,38 @@ def destination(self):
...
-        # MODIFIED: resource_types param added. None = full load (existing behavior).
-        self._data = self._storage.get(origin, resource_types=self._resource_types)
+        if self._exact_ids is not None:
+            # ID-targeted: fetch only specified resources by constructing keys directly
+            self._data = self._storage.get_by_ids(origin, self._exact_ids)
+        else:
+            # Type-scoped (resource_types set) or full load (resource_types=None)
+            self._data = self._storage.get(origin, resource_types=self._resource_types)
+
+    def ensure_resource_loaded(self, resource_type: str, resource_id: str) -> None:
```
michael-richey, line 94: This method has an implicit hard dependency on `resource_per_file=True` in the storage backend. `get_single` constructs paths like `{resource_type}.{id}.json` which only exist in per-file layout; with the legacy monolithic layout (`dashboards.json` containing all resources) it silently returns `(None, None)` for every call.

The CLI guard in `build_config` (`--minimize-reads requires --resource-per-file`) prevents this in production, but the method's contract doesn't document this constraint. Worth adding a note here so future callers don't get surprised:

```python
# Requires resource_per_file=True in the storage backend.
# get_single constructs per-resource filenames; monolithic layout
# will silently return (None, None) for every dependency.
```

datadog_sync/utils/configuration.py
```diff
@@ -135,6 +135,50 @@ async def exit_async(self):
...
+    for rt in resource_types:
+        rt_filters = filters.get(rt, [])
+        if not rt_filters:
+            return None  # No filters for this type — can't use ID-targeted
+        # All filters must be id-field ExactMatch
+        if not all(f.attr_name == ["id"] and f.operator == EXACT_MATCH_OPERATOR for f in rt_filters):
+            return None
+        # Extract raw IDs from ^...$-wrapped regex patterns
+        try:
+            ids = [_unwrap_exact_match_pattern(f.attr_re.pattern) for f in rt_filters]
```
michael-richey, line 177: The `ValueError` catch here is unreachable in practice. By the time we call `_unwrap_exact_match_pattern`, the outer `all(...f.operator == EXACT_MATCH_OPERATOR...)` guard has already confirmed every filter uses ExactMatch. `handle_deprecated_operator` always wraps ExactMatch values as `^...$`, so the pattern will always start with `^` and end with `$`. The fallback never fires.

If the intent is purely defensive, that's fine — but the docstring of `_unwrap_exact_match_pattern` says it raises `ValueError` 'so callers can gracefully fall back', implying this is the expected path. Either:
1. Remove the try/except and note that ExactMatch is guaranteed by the caller, or
2. Keep it but add a comment explaining this is dead-code-by-design defensiveness.

datadog_sync/utils/configuration.py
```diff
@@ -315,10 +359,18 @@ def build_config(cmd: Command, **kwargs: Optional[Any]) -> Configuration:
...
+    # Determine loading strategy for minimize-reads
+    _state_resource_types = None  # type-scoped; None = full load (existing behavior)
+    _state_exact_ids = None  # ID-targeted; None = not using ID-targeted
     if minimize_reads and (rs := kwargs.get("resources", None)):
-        _state_resource_types = [r.strip().lower() for r in rs.split(",") if r.strip()]
+        raw_types = [r.strip().lower() for r in rs.split(",") if r.strip()]
+        # Try ID-targeted strategy first (fast path: exact IDs from filters)
+        early_filters = process_filters(kwargs.get("filter"))
+        filter_operator = kwargs.get("filter_operator", "or")
+        _state_exact_ids = extract_exact_id_filters(early_filters, filter_operator, raw_types)
```
michael-richey, line 374: When ID-targeted loading falls back to type-scoped (or vice versa), there is only a single log line at INFO level recording the final strategy chosen. There is no signal to the operator explaining *why* the fast path was skipped.

In practice the common surprise will be: 'I passed exact ID filters, why is it doing 1000 reads instead of 100?' Adding a single debug/info log in `extract_exact_id_filters` (or here in `build_config` when `_state_exact_ids is None`) would make this diagnosable without resorting to source-level debugging:

```python
if _state_exact_ids is None:
    logger.debug(
        'minimize-reads: falling back to type-scoped — filters are not all id+ExactMatch+OR'
    )
    _state_resource_types = raw_types
```

datadog_sync/utils/resources_handler.py
```diff
@@ -625,6 +625,13 @@ def _resource_connections(self, resource_type: str, _id: str) -> Tuple[Set[Tuple
                         # After retrieving all of the failed connections, we check if
                         # the resources are imported. Otherwise append to missing with its type.
                         for f_id in failed:
+                            # With --minimize-reads, dependency types may not be in the
+                            # initial scoped load. Lazily load this specific dependency
+                            # (source+destination) so the source check below is accurate,
+                            # and so connect_resources() in _apply_resource_cb() can
+                            # successfully remap the ID in the destination.
+                            self.config.state.ensure_resource_loaded(resource_to_connect, f_id)
```
michael-richey, line 633: The PR description calls out a known limitation: `monitors.py` and `restriction_policies.py` have `connect_id()` overrides that access hardcoded destination types beyond what appears in `resource_connections`. Those types won't be in the scoped load and `ensure_resource_loaded` won't be called for them, so their IDs will silently not be remapped.

Silent ID-remapping failure is very hard to detect in production — the sync appears to succeed but references in the destination resource are stale. Even if fixing the root cause is deferred, the failure should be surfaced:

```python
# After ensure_resource_loaded + source check:
if f_id in self.config.state.source[resource_to_connect]:
    # it loaded successfully
else:
    if self.config.state._minimize_reads:
        log.warning(
            'minimize-reads: dependency %s.%s not found in storage; '
            'ID remapping may be incomplete', resource_to_connect, f_id
        )
    missing_resources.add((resource_to_connect, f_id))
```

At minimum, the test plan should include a case that syncs a monitor widget in a dashboard (which exercises the monitors override path) to confirm it doesn't silently break.

tests/unit/test_minimize_reads_id_targeted.py
```diff
@@ -0,0 +1,393 @@
...
+                    "aws_session_token": "",
+                },
+            )
+            src, dst = backend.get_single("dashboards", "nonexistent")
+            assert src is None
+            assert dst is None
+
+    def test_localfile_get_single_returns_none_for_missing_file(self, tmp_path):
+        """LocalFile get_single returns (None, None) when file doesn't exist."""
+        src_path = str(tmp_path / "source")
```
michael-richey, line 339: Missing coverage: repeated `ensure_resource_loaded` calls on an ID that was never stored (i.e., `get_single` always returns `(None, None)`). The current idempotency guard is:

```python
if resource_id in self._data.source[resource_type]:
    return
```

Because the resource is never inserted into source, this guard never triggers, and `get_single` is called on every invocation. Add a test like:

```python
def test_ensure_resource_loaded_repeated_miss_does_not_repeatedly_fetch(self, tmp_path):
    state = self._make_state_with_exact_ids(tmp_path)
    with patch.object(state._storage, 'get_single', return_value=(None, None)) as mock:
        state.ensure_resource_loaded('monitors', 'never-exists')
        state.ensure_resource_loaded('monitors', 'never-exists')
        state.ensure_resource_loaded('monitors', 'never-exists')
    # Should be called exactly once, not three times
    assert mock.call_count == 1
```

This test will fail today, which documents the bug described in the inline comment on `state.py:110`.

tests/unit/test_minimize_reads_id_targeted.py
```diff
@@ -0,0 +1,393 @@
...
+from datadog_sync.utils.storage.storage_types import StorageType
+
+
+# ─── Helpers ────────────────────────────────────────────────────────────────
+
+
+def make_exact_id_filter(resource_type: str, resource_id: str):
+    """Build a Filter object matching ExactMatch id=resource_id.
+
+    Mirrors process_filters() behavior: ExactMatch wraps value as ^...$
```
michael-richey, line 26: The `make_exact_id_filter` helper bypasses `process_filters` and constructs a `Filter` directly. That means if `process_filters` ever changes how ExactMatch patterns are constructed (e.g., adding `re.escape`), these tests won't catch the regression. Consider driving at least one test through `process_filters` directly so the integration between the filter parsing pipeline and `extract_exact_id_filters` is covered end-to-end.

datadog_sync/utils/state.py
michael-richey: Fixed. Added `self._ensure_attempted: set = set()` in `__init__` and updated `ensure_resource_loaded` to check/add to this set before the `get_single` call. The source-presence check is gone — the set now tracks *all* attempted lookups regardless of whether they were found.

datadog_sync/utils/configuration.py
michael-richey: Fixed. Added a comment above the try/except: 'Defensive: ExactMatch guarantees ^...$, so ValueError should not fire. Kept as a safety net in case filter construction changes upstream.' Also updated the `_unwrap_exact_match_pattern` docstring to remove the implication that this is the expected path.

tests/unit/test_minimize_reads_id_targeted.py
michael-richey: Fixed. Added `test_end_to_end_through_process_filters` to `TestExtractExactIdFilters`. It drives two ExactMatch filter strings through `process_filters` and then through `extract_exact_id_filters`, asserting the full pipeline produces the expected `{"dashboards": ["dash-1", "dash-2"]}` mapping.

tests/unit/test_minimize_reads_id_targeted.py
michael-richey: Fixed. Added `test_ensure_resource_loaded_repeated_miss_does_not_refetch` to `TestEnsureResourceLoaded`. Uses `patch.object` on `get_single` returning `(None, None)` and asserts `call_count == 1` after three invocations with the same ID.

datadog_sync/utils/resources_handler.py
michael-richey: Fixed. Added a `self.config.logger.warning` call after the source check in `_resource_connections`. Noted in the comment that the monitors.py override (which accesses synthetics_tests directly in connect_id) won't be caught here since that path bypasses \_resource_connections — fixing the root cause is a separate refactor.

datadog_sync/utils/configuration.py
michael-richey: Fixed. Added a debug log in the fallback branch: `logger.debug('minimize-reads: ID-targeted not eligible — filters are not all id+ExactMatch+OR')`.

datadog_sync/utils/state.py
michael-richey: https://github.com/DataDog/datadog-sync-cli/pull/547

riyazsh (approved):

riyazsh (dismissed):

riyazsh (dismissed):

riyazsh (approved):

PR #560, opened 2026-05-13
feat(import): add ImportState + --skip-state-load flag for write-only import

riyazsh:
## Summary

`State.__init__` unconditionally loads existing state from storage. The `import` command discards prior per-type source state before its API fetch (in `_import_get_resources_cb`) and writes fresh data via `dump_state` at the end — the boot-time load is dead weight for that command. On populated buckets the load can dominate per-run wall-clock.

This PR adds an opt-in `--skip-state-load` flag to `import` that constructs a new `ImportState` instead of `State`. `ImportState` exposes only write methods and has no `source` / `destination` property, so any attempt to read state fails at the language level with `AttributeError` rather than silently returning empty data.

Adds diagnostic timing logs around state init, load, dump, and each storage backend's `_list_and_load` and `put`. The logs emit from a `finally` block so they fire even when an SDK call raises mid-iteration, and carry a `list_ms` vs `download_ms` split plus pagination and blob counts. Useful for diagnosing where wall-clock goes on populated buckets.

## What changes

- New `ImportState` class (`utils/import_state.py`) — write-only state for the import command. `set_source`, `clear_source_type`, `dump_state(Origin.SOURCE)`. No read accessors; `dump_state` rejects DESTINATION / ALL.
- New `SourceStateWriter` Protocol (`utils/source_state_writer.py`) — shared write surface implemented by both `State` and `ImportState`.
- New `--skip-state-load` Click option on `import` only. Click rejects the flag on `sync` / `migrate` / `diffs` / `reset` / `prune`; `build_config` adds a redundant validation guard.
- Three import-path writer sites migrated to the protocol methods: `base_resource._import_resource`, `resources_handler._import_get_resources_cb`, `model/synthetics_private_locations.import_resource`.
- Shared `build_storage_backend` helper in `_base_storage.py` so `State` and `ImportState` cannot drift on per-backend init.
- Diagnostic timing instrumentation across `state.py` and all four storage backends, with a `try/finally` so the log emits even when the underlying SDK raises. New fields: `pages_listed`, `blobs_listed`, `blobs_downloaded`, `transient_errors`, `aborted`, `list_ms`, `download_ms`, `wall_ms`. Race-delete semantics (`NoSuchKey` on AWS, `ResourceNotFoundError` on Azure, `NotFound` on GCS) are caught uniformly across backends and counted in `transient_errors`.

## What does not change

- Default behavior. `--skip-state-load` is opt-in (`default=False`); existing callers see no behavior change.
- `sync` / `diffs` / `migrate` / `reset` / `prune` paths. These continue to use `State` with the existing read+write surface.
- Wire format, file layout on disk, or storage backend semantics. `dump_state` patches per-blob exactly as before.

## Constraints

`--skip-state-load` requires `--resource-per-file` and `--resources`, and is not combinable with `--minimize-reads` (which serves a different optimization goal). Validated in `build_config`.

## Test plan

- [x] 556 unit tests pass (514 existing + 42 new in `test_import_state.py`)
- [x] `black --line-length 120` clean
- [x] New coverage: ImportState invariant (no read accessors), write API, Protocol satisfaction by both classes, partial-import preserves untouched types end-to-end, stale-within-type unchanged from baseline, CLI validation errors, Click rejection on all non-import commands, deprecated-resource conflict still rejected with `--skip-state-load`, timing-log schema fields present on all four backends, **aborted=1 fires when each backend's SDK call raises**, paged-iterator branch on GCS, round-trip read of `--skip-state-load`-written state via a fresh `State`.

## Rollout

Flag is opt-in (`default=False`). Reverting the commit removes the option from `import --help` and doesn't break any caller that wasn't using it. No on-disk or wire-format effect.

## Out of scope

- `migrate` does not gain `--skip-state-load` (apply-phase reads destination).
- `sync` keeps its existing state-load path.
- Parallelising `_list_and_load` is a separate effort.

### DIFF
--- a/datadog_sync/commands/_import.py
+++ b/datadog_sync/commands/_import.py
@@ -3,9 +3,10 @@
 # This product includes software developed at Datadog (https://www.datadoghq.com/).
 # Copyright 2019 Datadog, Inc.
 
-from click import command
+from click import command, option
 
 from datadog_sync.commands.shared.options import (
+    CustomOptionClass,
     common_options,
     destination_auth_options,
     force_missing_dependencies_options,
@@ -22,6 +23,19 @@ from datadog_sync.constants import Command
 @common_options
 @force_missing_dependencies_options
 @storage_options
+@option(
+    "--skip-state-load",
+    required=False,
+    is_flag=True,
+    default=False,
+    show_default=True,
+    help="Skip loading prior state from storage. Import discards prior source "
+    "state per type and writes fresh API results, so the boot-time read is "
+    "dead weight; this flag lets you reclaim that cost on populated buckets. "
+    "Requires --resource-per-file and --resources. Not available on other "
+    "commands.",
+    cls=CustomOptionClass,
+)
 def _import(**kwargs):
     """Import Datadog resources."""
     run_cmd(Command.IMPORT, **kwargs)
--- a/datadog_sync/model/synthetics_private_locations.py
+++ b/datadog_sync/model/synthetics_private_locations.py
@@ -49,7 +49,7 @@ class SyntheticsPrivateLocations(BaseResource):
             raise SkipResource(import_id, self.resource_type, "Managed location.")
 
         pl = await source_client.get(self.resource_config.base_path + f"/{import_id}")
-        self.config.state.source[self.resource_type][import_id] = pl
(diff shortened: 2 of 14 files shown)


heyronhay (approved):
Nothing major that I can find, looks good!

PR #567, opened 2026-05-18
feat(notebooks): lightweight LIST + per-id GET

riyazsh:
LIST notebooks without include_cells and fetch each notebook body via GET /api/v1/notebooks/{id}. Per-page response size with cells included grows with cell count and dominates import wall-clock on populated orgs. Matches the dashboards pattern.

Folded-in correctness items so behavior is not regressed vs the prior include-cells-on-LIST contract:

* 403/404 on the per-id GET raise SkipResource (restricted notebook or deleted-between-list-and-fetch) instead of failing the run.

* New ResourceConfig.list_omitted_attr_prefixes, set on Notebooks (attributes.cells) and Dashboards (widgets). The handler partitions the user's --filter at LIST-time: list-safe filters (e.g. attributes.name) still short-circuit per-page; list-unsafe filters (those referencing an omitted prefix) defer to a new post-GET pass in base_resource._import_resource which raises FilteredResource on rejection. Without this, positive filters on attributes.cells.X / widgets.X silently no-op against the metadata-only LIST item. AND/OR decisiveness preserved: AND-miss is decisive reject; OR-miss with deferred clauses defers to GET so a deferred match can rescue.

* --force-missing-dependencies bypasses the post-GET filter via a skip_filter kwarg. Force-imported deps must be present in source state regardless of --filter, or run_sorter() will hit unresolved references on the dependent resource.

* Both models short-circuit import_resource when the caller already supplied a body (cells / widgets present). The --id-file path's get_resources_by_ids prefetches bodies; without this, the queue handler's second pass would re-GET each resource. Pre-existing latent double-fetch in dashboards; addressed for both models.

## Test plan

- [x] 600 unit tests pass, covering the LIST contract, per-id GET, prefetched-body short-circuit, 403/404 SkipResource, post-GET filter accept/reject, AND/OR decisiveness for mixed list-safe/list-unsafe filters, and --force-missing-dependencies filter bypass.
- [ ] Integration cassettes under tests/integration/resources/cassettes/test_notebooks/ and test_restriction_policies/ need to be re-recorded against the new request shape before the integration workflow is enabled on this PR.


heyronhay (approved):

heyronhay (dismissed):
Looks good!

michael-richey (approved):

PR #593, opened 2026-06-09
[CHUN-12] feat(team_memberships): team_memberships fan-out + _ID_FILE_SUPPORTED_TYPES (PR 7)

michael-richey:
## Summary

Fixes the broken 1-arg `import_resource(team_uuid)` path and enables team_memberships for `--id-file`.

## Problem

`import_resource(_id)` (1-arg path) called `paginated_request()` which returns a list. The code then used `cast(dict, resource)` (a Python typing no-op) and called `resource.get("id")` — raising `AttributeError: 'list' object has no attribute 'get'` at runtime.

## Solution

**Dispatch by `:` presence:** `":" not in _id` distinguishes bare team UUID (from `--id-file`) from composite `team_id:user_id` (from 2-arg `get_resources()` path).

**1-to-N fan-out:** For bare team UUID, fetch `/api/v2/team/{uuid}/memberships` and write N composite rows `f"{team_id}:{user_id}"`.

**Delete-before-write:** Uses new `delete_source()` API (not direct dict mutation) to remove stale `team_uuid:*` rows before writing refreshed membership set. This ensures users who left the team are not retained. Required because `set_source` is overwrite-per-exact-key only (P3 verification in CHUN-5).

## New Public Methods

- `ImportState.delete_source(resource_type, key)` — remove stale state rows
- `ImportState.get_source_keys(resource_type)` — list keys for delete-before-write  
- Same methods added to `State` for consistency

## No Feature Flag

Fan-out is unconditional once this PR merges. The upstream type-enable configuration is the sole gate — operators must not enable `team_memberships` until this PR is pinned in the orchestrator.

## Merge Order

**PR 7 must NOT merge until PR 6 (`authn_mappings`) is merged and pinned.** Both PRs may be reviewed in parallel.

## Test Coverage

11 TDD tests covering: fan-out count, composite key format, empty team, 2-arg path unaffected, API error propagation, pagination, idempotency, cross-team users, member removal, overwrite-not-append invariant.

## Jira

Resolves CHUN-12. Part of CHUN-1 (epic).

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### Update 2026-06-09 (CI follow-up)
- Replaced `asyncio.get_event_loop().run_until_complete(...)` with `asyncio.run(...)` across `tests/unit/test_team_memberships_id_file.py` for Python 3.12 compatibility.
- Removed unused `copy` import in that test module.
- Validation: `tox -e py311 -- tests/unit/test_team_memberships_id_file.py` (11 passed).
- Fix commit: `89875527`.

### DIFF
--- a/datadog_sync/model/team_memberships.py
+++ b/datadog_sync/model/team_memberships.py
@@ -78,22 +78,13 @@ class TeamMemberships(BaseResource):
         return all_team_memberships
 
     async def import_resource(self, _id: Optional[str] = None, resource: Optional[Dict] = None) -> Tuple[str, Dict]:
-        source_client = self.config.source_client
-        pagination_config = PaginationConfig(
-            page_size=100,
-            page_number_param="page[number]",
-            page_size_param="page[size]",
-            remaining_func=lambda idx, resp, page_size, page_number: max(
-                0,
-                resp["meta"]["pagination"]["total"] - (page_size * (idx + 1)),
-            ),
-        )
-
-        if _id:
-            resource = await source_client.paginated_request(source_client.get)(
-                self.team_memberships_path.format(_id),
-                pagination_config=pagination_config,
-            )
+        if _id is not None and resource is None:
+            if ":" not in _id:
+                # _id is a bare team UUID from --id-file; fan out to N membership rows.
+                # Dispatch condition: ":" not in _id distinguishes bare team UUID
+                # (from --id-file) from composite team_id:user_id (from 2-arg path).
+                return await self._import_team_memberships_by_team_id(_id)
+            # composite team_id:user_id from get_resources() falls through to existing code
 
         resource = cast(dict, resource)
         if not resource:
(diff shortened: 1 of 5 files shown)


michael-richey:
Review pass (PR7):

- Fan-out + delete-before-write flow is largely in line with the design and the test coverage is strong.
- Potential behavior issue: the empty-team path returns `(team_id, {})`, which causes the outer `_import_resource` write to persist a bare `team_id` key. That is not a composite membership row and may linger in source state.
- Can we avoid writing that marker row for empty fan-outs? If intentional, please document why it is safe long-term in persisted state.

michael-richey:
Fixed in 03092515.

**Root cause confirmed:** returning `(team_id, {})` for empty fan-outs caused the outer `_import_resource` to call `set_source(team_id, {})`, writing a bare `team_id` key with no colon. The delete-before-write prefix match (`team_id:`) would never catch that key, so it would linger permanently.

**Changes made:**
- `_import_team_memberships_by_team_id`: raise `SkipResource("team has no members")` for empty teams instead of returning the bare key. `SkipResource` is caught by `fetch_one` in the batch import path and treated as a non-error skip — no `set_source` call is made.
- Delete-before-write: extended to also delete the bare `team_id` key (`k == team_id`) in addition to `team_id:`-prefixed keys, so stale bare keys from prior runs of the old code are cleaned up.
- Test updated to `pytest.raises(SkipResource)` and asserts zero total state keys (not just zero composite keys).
- Removed unused `patch` import (pre-empts ruff F401).

michael-richey:
Re-reviewing the latest update (`0309251`) — the functional behavior from the previous blocker looks fixed:

- delete-before-write + `SkipResource` for empty teams addresses the stale/bare-key state issue.

I still see one blocker to resolve before merge:

- New async tests in `tests/unit/test_team_memberships_id_file.py` repeatedly use `asyncio.get_event_loop().run_until_complete(...)`.
- Under Python 3.12 this pattern raises `RuntimeError: There is no current event loop in thread 'MainThread'` unless a loop is created explicitly.

Please switch those calls to `asyncio.run(...)` (or make them `pytest.mark.asyncio` tests and `await` directly).

Minor nit while touching tests: `import copy` appears unused in that file.

michael-richey:
Addressed the CI blocker from re-review:

- Updated async test execution in `tests/unit/test_team_memberships_id_file.py` to use `asyncio.run(...)` throughout.
- Removed the unused `copy` import in that file.
- Pushed in `89875527`.
- Local check: `tox -e py311 -- tests/unit/test_team_memberships_id_file.py` (11 passed).

Ready for CI re-run.

michael-richey:
Superseded by consolidated replacement PR: #594\n\nClosing this stacked PR in favor of the single replacement for simpler review and merge.

michael-richey:
Closing as superseded by consolidated replacement PR: #594.

PR #594, opened 2026-06-09
feat(id-file): enable authn_mappings + team_memberships

michael-richey:
## What
- add `authn_mappings` and `team_memberships` to `_ID_FILE_SUPPORTED_TYPES`
- implement `team_memberships` id-file fan-out in `get_resources_by_ids(...)` (`team_id` -> N membership resources)
- keep the id-file fetch phase side-effect free for `team_memberships` (no source-state mutation during fetch)
- reject direct `team_memberships.import_resource(_id=...)` usage; fan-out is handled by `get_resources_by_ids(...)`
- remove now-unused source-state helper methods that were added for fetch-time mutation
- include focused unit tests for both resource types
- use `asyncio.run(...)` in new tests for Python 3.12 compatibility

## Why
- ease batching these resources
- preserve second-pass import and filter semantics for id-file runs

## Validation
- `pytest tests/unit/test_team_memberships_id_file.py tests/unit/test_authn_mappings_id_file.py tests/unit/test_get_resources_by_ids_experiment.py::test_id_file_allowlist_rejects_non_monitors_v1 tests/unit/test_get_resources_by_ids_experiment.py::test_id_file_allowlist_accepts_monitors`

### DIFF
--- a/datadog_sync/commands/shared/options.py
+++ b/datadog_sync/commands/shared/options.py
@@ -184,14 +184,14 @@ _common_options = [
         help="Max number of workers when running operations in multi-threads.",
         cls=CustomOptionClass,
     ),
-    # ID-targeted import flags (monitors only in v1)
+    # ID-targeted import flags (supported allowlisted types only)
     option(
         "--id-file",
         required=False,
         default=None,
         help="Path to JSON file mapping resource types to ID lists, or `-` for stdin. "
         "When set, import command fetches only the specified IDs instead of "
-        "listing all resources. v1 supports monitors only.",
+        "listing all resources. Supported types are enforced by a code-level allowlist.",
         cls=CustomOptionClass,
     ),
     option(
--- a/datadog_sync/model/roles.py
+++ b/datadog_sync/model/roles.py
@@ -398,9 +398,7 @@ class Roles(BaseResource):
         # Reverse map: destination UUID -> permission name (for sync path where source
         # state still holds permission names after import_resource's name rewrite).
         uuid_to_name = (
-            {uuid: name for name, uuid in self.destination_permissions.items()}
-            if self.destination_permissions
-            else {}
+            {uuid: name for name, uuid in self.destination_permissions.items()} if self.destination_permissions else {}
         )
         # A permission entry in source_perms may carry either a name (sync path, pre-remap)
         # or a destination UUID (diffs path, where pre_resource_action_hook just ran on the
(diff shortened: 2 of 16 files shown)


michael-richey:
Rationale for `team_memberships` id-file fan-out (`team_id -> team_id:user_id`):

- `--id-file` provides a **team id**, but team memberships are naturally a **set of rows** keyed by `(team_id, user_id)`.
- The source API is team-scoped (`/api/v2/team/{team_id}/memberships`), so we must fetch members for that team and expand to per-membership records.
- State/mapping for this resource already uses composite keys (`team_id:user_id`) via `_team_membership_key`, so fan-out preserves existing identity semantics and diff behavior.
- The implementation does delete-before-write for that team prefix to remove stale memberships for users no longer in the team.
- Empty team case intentionally raises `SkipResource` so we don't persist an invalid bare `team_id` row in state.

riyazsh:
**Codex Findings**
**MAJOR** - [datadog_sync/model/team_memberships.py#L102](https://github.com/DataDog/datadog-sync-cli/blob/08fdb963ee28173d981633562e490e2045300d90/datadog_sync/model/team_memberships.py#L102): the hand-rolled targeted pagination only stops when len(members_of_team) < page_size, ignoring meta.pagination.total. Reproducer: team has exactly 100 memberships, page 0 returns 100 members with total=100, then the unnecessary page 1 request hits 429/5xx. The code classifies the whole team as transient and drops the 100 complete members. Fix: stop when the fetched count reaches meta.pagination.total or reuse the existing remaining_func; add a regression test that page 1 is not called when total == page_size.
**MINOR** - [datadog_sync/model/team_memberships.py#L150-L151](https://github.com/DataDog/datadog-sync-cli/blob/08fdb963ee28173d981633562e490e2045300d90/datadog_sync/model/team_memberships.py#L150-L151): a valid empty team is reported as skipped/errored even though the method contract says a team ID fans out to “zero or more” memberships. Reproducer: /memberships returns {"data": []} for an existing team; the import emits a skipped outcome for a successful no-op. Fix: return success with an empty list, or emit a distinct success/noop outcome if callers need visibility.
**NIT** - [datadog_sync/utils/configuration.py](https://github.com/DataDog/datadog-sync-cli/blob/08fdb963ee28173d981633562e490e2045300d90/datadog_sync/utils/configuration.py): the _ID_FILE_SUPPORTED_TYPES docstring still says “Monitors only in v1” after adding authn_mappings and team_memberships. Update that and the stale test wording around “non-monitors” to avoid future reviewer confusion.

heyronhay (approved):

riyazsh (approved):

PR #601, opened 2026-06-30
Update doc for synthetics-private-locations

melkouri:
### What does this PR do?

Adds documentation for using Synthetics Private Locations with the Disaster Recovery (DDR) setup.

### Description of the Change

Adds `docs/synthetics-private-locations.md`, a guide explaining how to replicate Synthetics Private Locations from a source to a destination org using `datadog-sync-cli`, and how to configure the PL worker's `datadogHostOverride` for failover. Updates the README to list `synthetics_private_locations` as a supported resource with a link to the new guide.

### DIFF
--- a/docs/synthetics-private-locations.md
+++ b/docs/synthetics-private-locations.md
@@ -9,7 +9,6 @@ that test traffic continues flowing when the destination region takes over (fail
 - [Overview](#overview)
 - [Replicating PLs with `datadog-sync-cli`](#replicating-pls-with-datadog-sync-cli)
   - [What gets synced](#what-gets-synced)
-  - [The `--datadog-host-override` flag](#the---datadog-host-override-flag)
 - [How replicated PLs work after sync](#how-replicated-pls-work-after-sync)
 - [Configuring `datadogHostOverride` on your PL workers](#configuring-datadoghostoverride-on-your-pl-workers)
   - [Option A: env var (Kubernetes / Helm chart)](#option-a-env-var-kubernetes--helm-chart)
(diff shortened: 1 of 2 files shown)


docs/synthetics-private-locations.md
```diff
@@ -0,0 +1,198 @@
...
+
+In practice, pick one mechanism and stick with it for clarity.
+
+## Verifying the setup
+
+After running the sync and starting your worker with `datadogHostOverride` set:
+
+1. **Confirm the PL appears in both orgs.** Check the Synthetics → Settings → Private
+   Locations page in each org. The PL should be listed with the same name in both.
+2. **Confirm the worker reports.** The PL should show as healthy in both orgs once the
```
aletournel: I don't agree with it because a PL can only be reporting to main DCs (and so one org) at a time. During the exercice only the DC where the override CNAME was point to, had an "healthy" PL

docs/synthetics-private-locations.md
```diff
@@ -0,0 +1,198 @@
...
+   worker starts polling. If it only reports in one, double-check that the CNAME's current
+   DNS target is the region you expect.
+3. **Run a test.** Create a test in either org targeting the replicated PL, unpause it,
+   and confirm results appear. During steady state, results land in your primary region;
```
aletournel: Same the result will appear only in the org/main DC that the CNAME used in datadogHostOverride parameter is resolving to.

docs/synthetics-private-locations.md
```diff
@@ -0,0 +1,198 @@
...
+```bash
+--datadog-host-override <your-failover-cname>.synthetics.datadoghq.com
+```
+
+That CNAME gets stored as `datadogHostOverride` in the destination PL's config, which the
+worker then picks up.
+
+If you don't pass the flag, the destination PL gets created without a `datadogHostOverride`,
+which means the worker will fall back to the standard region intake for the secondary org.
+That's usually **not** what you want for a failover setup — pass the flag.
```
aletournel: As discussed this part maybe simplified or removed as we probably don't need to take care of the datadogHostOverride during the sync. 

The next section `Configuring datadogHostOverride on your PL workers` is what actually matter

aletournel (approved):
LGTM 👍

heyronhay (approved):

PR #616, opened 2026-07-09
Port restriction_policy remap to synthetics_tests

riyazsh:
## Summary

Ports the full `restriction_policy` principal-remap pattern from `monitors.py` to `synthetics_tests.py`, closing two related failure modes on DDR sync.

## Fixes

- **`org:` cross-org rejection** — API rejects when the binding contains an `org:` principal pointing at the source org UUID (`cross-org principals are not supported`).
- **`user:`/`role:`/`team:` cross-org rejection** — API rejects when principals reference source-org IDs not on destination (`one or more roles/users/teams do not exist in organization`).

## Change

- `pre_apply_hook` fetches destination org UUID via `GET /api/v2/current_user`; stores as `self.org_principal`.
- `pre_resource_action_hook` remaps any `org:` principal to `self.org_principal`.
- `resource_connections` wires `restriction_policy.bindings.principals` to `roles`, `users`, `teams`.
- `connect_id` + `extract_source_ids` add a `key == "principals"` branch for prefixed remap.
- 15 unit tests mirroring `test_monitors.py`.

All 25 tests in the file pass; ruff clean.

## Verified

`synthetics_mobile_applications*.py` do not carry `restriction_policy` (grep-verified) — no port needed.

Ref: HAMR-392 Jul8-T15. Diagnosis at `team-compass/teams/hamr-ddr/tickets/active/HAMR-392/T-synthetics-restriction-policy/STATUS.md@ede7b066ee`.


nathantournant (approved):
LGTM — clean port of the restriction_policy principal-remap pattern from monitors.py, and the unit coverage mirrors it well.

One thing worth a potential follow-up (non-blocking): `pre_apply_hook` now issues `GET /api/v2/current_user` and re-raises on failure, which means the entire synthetics apply phase becomes contingent on that call succeeding — even for tests that carry no restriction_policy. On destinations where credentials lack current_user access (or on a transient failure / unexpected response shape → KeyError), this would abort the whole synthetics sync rather than just the affected resources. Since it mirrors monitors.py this is consistent behavior, but it might be worth scoping the org lookup to only run when at least one resource actually has a restriction_policy, so policy-free tests don't gain a new failure dependency.

heyronhay (approved):

riyazsh:
Thanks @nathantournant! Agree the optimization is worth doing. I'll ship #616 as-is (matches the existing `monitors.py` shape so behavior stays consistent), and file a follow-up PR that scopes the `current_user` lookup to only run when at least one resource in the batch actually carries a `restriction_policy`. Best done as a shared helper on `BaseResource` so `monitors.py`, `synthetics_tests.py`, and `restriction_policies.py` all benefit at once.

riyazsh:
Follow-up filed: #622 — implements the scoped `current_user` lookup as a shared `BaseResource` helper across monitors, synthetics_tests, and restriction_policies. Thanks again @nathantournant.

PR #622, opened 2026-07-09
Scope current_user lookup to policy-carrying resources

riyazsh:
## Summary

Follow-up to #616 addressing @nathantournant's non-blocking review comment: only fetch destination org UUID when at least one source resource actually carries a `restriction_policy`.

## Change

- New helper `BaseResource._fetch_destination_org_principal(has_policy)` scans `state.source[resource_type]` and calls `GET /api/v2/current_user` only when the predicate matches.
- `monitors.py`, `synthetics_tests.py`, `restriction_policies.py` `pre_apply_hook` now delegate to the helper. Behavior unchanged when policies exist; skipped entirely otherwise.
- 4 new unit tests per resource covering: success, failure, empty-source-skip, no-policy-skip.
- Integration cassettes re-recorded (monitors + synthetics_tests).

## Motivation

@nathantournant: *"On destinations where credentials lack current_user access (or on a transient failure / unexpected response shape → KeyError), this would abort the whole synthetics sync rather than just the affected resources."*

After this PR: syncs that carry no `restriction_policy` don't touch `/api/v2/current_user` at all — no unnecessary call, no new failure dependency.

Ref: HAMR-392


nathantournant (approved):
Approving — the scoped current_user lookup is a solid reliability fix (no more aborting policy-free syncs on a current_user permission/shape failure), and the gate correctly matches the remap-consumption condition. Pushed a small follow-up to seed source state in test_restriction_policies.py so the current_user GET (and its failure path) is actually exercised; CI is now green across all jobs.

riyazsh:
Thanks @nathantournant — much appreciated for the test-state seed fix. That's the exact patch I was about to push based on a Codex review flagging the same P1. Pulled your commit locally.

PR #626, opened 2026-07-14
fix(downtime_schedules): rewrite past `end` on create path

riyazsh:
## Summary

`pre_resource_action_hook` already rewrites past `schedule.start` forward on the create path. This extends the create-path handling to `schedule.end`:

- **Past `end` → SkipResource.** A downtime whose end time has already passed is a closed maintenance window with nothing left to silence. Replicating it either fails with "Downtime cannot be scheduled in the past" or invents a phantom window on the destination that never existed on the source.
- **Past `start` with future `end` → bump `start`, leave `end` as-is.** The customer's intended end time is preserved; the window may shrink.
- **Both future → no-op** (existing behavior).

## Also

Create-path timestamp comparisons hardened to UTC-aware `datetime` so `.timestamp()` is correct on non-UTC hosts. Prior code compared naive `utcnow()` against tz-aware `parse()` results; correct on UTC hosts, drifted by TZ offset elsewhere.

Update path (backward-clamp of source to destination stored values) is intentionally out of scope; a follow-up will address the update-path case.

## Test plan

- [x] 8 new unit tests covering: past-start + no end (bumped), past-start + future-end (start bumped, end preserved), past-end + past-start (skipped), past-end + future-start (skipped), both future (untouched), missing/null schedule (no-op), start-only no-end field (bumped, end field not injected), update-path untouched.
- [x] Full unit suite green apart from 6 pre-existing failures in `test_custom_client_trust_env.py` (unrelated).
- [x] Pre-merge review applied (two independent prompts, structural + checklist). Findings addressed across commits.

### DIFF
--- a/datadog_sync/model/downtime_schedules.py
+++ b/datadog_sync/model/downtime_schedules.py
@@ -84,26 +84,38 @@ class DowntimeSchedules(BaseResource):
             now = datetime.now(timezone.utc)
             floor = now + timedelta(seconds=60)
 
-            # Rewrite past `start` forward. Existing behavior; parse/format hardened
-            # to UTC-aware so `.timestamp()` is correct on non-UTC hosts.
+            # Rewrite past `start` forward. Existing behavior; parse/format
+            # hardened to UTC-aware so `.timestamp()` is correct on non-UTC hosts.
             start_raw = schedule.get("start")
-            start_dt = None
-            if start_raw:
-                start_dt = self._parse_utc(start_raw)
-                if start_dt <= now:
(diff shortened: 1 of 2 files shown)


heyronhay (commented):
We should never move the end time for a downtime, so we shouldn't do this PR ... I think.

riyazsh:
> We should never move the end time for a downtime, so we shouldn't do this PR ... I think.

Thanks for the catch. Updated PR.

michael-richey (approved):

PR #636, opened 2026-07-17
feat(users): create users via the v1 API when --use-v1-user-api is set

michael-richey:
## What
When `--use-v1-user-api` is set, create users via `POST /api/v1/user` (which accepts an explicit handle) instead of v2.

## Why
The destination derives a user's handle from its email on v2 create. When two source users share an email but have distinct handles, v2 create collapses them: the first create silently takes a handle that may belong to another user, and the second create 409s and can never succeed. v1 accepts an explicit handle, so each user is created with its own handle and no create collides.

## Changes
- With the flag set, `create_resource` posts to `/api/v1/user` with the source handle; with the flag off, user creation is unchanged (v2).
- The v1 response is the legacy user shape, so re-resolve the v2 UUID by exact-case handle (small, sleep-free re-query loop for read-after-write visibility) and return that record — state and downstream references (roles, team memberships) key on the v2 UUID.
- Any failure re-raises so the apply loop records it and continues with the other resources (disaster-recovery behavior preserved).

## Tests
`tests/unit/test_users.py`: flag off → v2 path only; flag on → v1 POST with the handle, v2 never called, UUID reconciled; **two shared-email/distinct-handle users are both created via v1 with their own handles** (the core fix); empty-handle guard; v1-POST failure re-raises; reconcile re-queries on an empty first page and selects the exact-case handle among candidates; existing-handle update path takes no create.

Stacked on #635 (part 2 of 3).


michael-richey:
Consolidated into #635 — the stack was collapsed into a single PR.

PR #637, opened 2026-07-17
docs(users): document --use-v1-user-api

michael-richey:
## What
Document the opt-in `--use-v1-user-api` flag: a README section (mirroring the existing flag sections) and a table-of-contents entry.

## Why
Completes the stack with user-facing docs for the v1 user-creation path.

Stacked on #636 (part 3 of 3). Docs-only.


michael-richey:
Consolidated into #635 — the stack was collapsed into a single PR.

PR #639, opened 2026-07-24
Add opt-in v2 user handle passthrough

michael-richey:
## Stack

This is the first PR in the user-creation cleanup stack. #640 is stacked on this PR and removes the legacy v1 creation path after this replacement is available.

## Summary

Add an opt-in `--preserve-user-handle` flag and `DD_PRESERVE_USER_HANDLE` environment variable to the `sync` and `migrate` commands.

When enabled, v2 user creation passes a non-empty source `attributes.handle` through unchanged to `POST /api/v2/users`. The option defaults to `false`, so existing behavior is unchanged.

## Behavior

- With the option disabled, v2 create continues to omit `handle`.
- With the option enabled and a non-empty source handle, v2 create includes that handle.
- With the option enabled but no source handle, v2 create still omits the field so the destination can derive it.
- V2 updates always omit `handle`; this option does not rename existing mapped users.
- Handle-only differences remain excluded from update diffs.
- V2 create errors propagate through the existing per-resource failure boundary so remaining resources can continue processing.
- Debug logging records whether the option is enabled without logging the handle value.

## Compatibility and limitations

- This PR leaves `--use-v1-user-api` and the existing v1 creation path unchanged; #640 removes them.
- The new option is registered for `sync` and `migrate`, not `import`.
- The destination must support explicit handles during v2 user creation. Otherwise, the destination API may reject the request.
- Handle validation remains the destination API's responsibility.

The README documents the option and its net-new-user limitation.

## Test plan

- [x] Enabled, disabled, and missing-handle v2 create payloads
- [x] V2 update payload and existing-user behavior
- [x] CLI wiring for `sync` and `migrate`
- [x] Configuration defaults and help output
- [x] Create-error propagation and nominal enabled/disabled paths
- [x] Full Python 3.11 unit suite


michael-richey:
Closing — superseded. This branch ended up implementing a different, unreviewed approach (an internal SAML NameID feature flag has broader effects on SSO login/JIT-provisioning behavior beyond just user creation, which this PR didn't account for). Not pursuing this direction. #640 will be reworked to stand alone (removing --use-v1-user-api directly from main) rather than stacking on this.

PR #649, opened 2026-07-28
fix(slo_corrections): strip null-valued mutually-exclusive fields before POST

riyazsh:
## Summary

Destination `POST /api/v1/slo/correction` rejects payloads that carry `slo_id`, `slo_query`, or `end` explicitly set to `null`. Error shape:

```
API input validation failed: {'slo_query': ['Field may not be null.']}
```

Source resources round-tripped from `GET /api/v1/slo/correction/{id}` carry both `slo_id` and `slo_query` with one of them explicitly null — they are mutually exclusive:

- **Time-based correction**: `slo_id` set, `slo_query = null`
- **Query-based correction**: `slo_query` set, `slo_id = null`

Similarly, recurring rrule corrections carry `end = null` (no fixed end time).

## Change

Add `attributes.slo_id`, `attributes.slo_query`, and `attributes.end` to `SLOCorrections.resource_config.non_nullable_attr`. The existing `del_null_attr` helper (`datadog_sync/utils/resource_utils.py:277`) strips fields listed there when their value is `None`, via `prep_resource`. Same pattern already used for `attributes.duration` and `attributes.rrule` (mutually-exclusive with each other), and for `downtimes.recurrence.until_date` vs `.until_occurrences`.

## Safety

- **Symmetric strip**: `resources_handler.py:503` (resource) + `:506` (destination_copy) both pass through `prep_resource` before `check_diff` — no spurious diff on state-vs-resource comparison when one side has `null` and the other has the field stripped.
- **Order with `connect_resources`**: `connect_resources` runs at `resources_handler.py:496`, before `prep_resource` at `:503`. Populated `slo_id` gets remapped first; only the null `slo_id` (the mutually-exclusive case) is stripped.
- **`is None` semantics**: `del_null_attr` at `resource_utils.py:285` uses `resource[k_list[0]] is None`, so integer `0` and empty string are preserved (matches destination API semantics, since 0 is a valid unix timestamp for `end`).

## Tests

New file `tests/unit/test_slo_corrections.py` with 6 unit tests:

1. `test_time_based_correction_strips_null_slo_query` — populated `slo_id`, null `slo_query` → `slo_query` stripped, `slo_id` + `end` preserved
2. `test_query_based_correction_strips_null_slo_id_and_null_end` — populated `slo_query`, null `slo_id` + null `end` → both stripped
3. `test_populated_fields_are_preserved` — nothing null → nothing stripped
4. `test_zero_end_is_preserved_not_stripped` — `end: 0` (integer zero) is NOT null and must survive
5. `test_check_diff_symmetric_strip_no_spurious_diff` — symmetric `prep_resource` on both sides of `check_diff` yields empty diff; regression pin against a future asymmetric-prep refactor
6. `test_null_duration_and_rrule_still_stripped_regression` — existing `duration` + `rrule` behavior preserved

Local pytest run: 6/6 pass. Broader unit suite: 1067 pass, only pre-existing unrelated failures in `test_custom_client_trust_env.py` (present on unmodified `main`).

## Rollback

Three-line revert of the `non_nullable_attr` list additions.

### DIFF
--- a/datadog_sync/model/slo_corrections.py
+++ b/datadog_sync/model/slo_corrections.py
@@ -20,7 +20,21 @@ class SLOCorrections(BaseResource):
         resource_connections={"service_level_objectives": ["attributes.slo_id"]},
         base_path="/api/v1/slo/correction",
         excluded_attributes=["id", "attributes.creator", "attributes.created_at", "attributes.modified_at"],
-        non_nullable_attr=["attributes.duration", "attributes.rrule"],
+        non_nullable_attr=[
+            "attributes.duration",
+            "attributes.rrule",
+            # slo_id and slo_query are mutually exclusive: a time-based
+            # correction sets slo_id (with slo_query=null) and a
+            # query-based correction sets slo_query (with slo_id=null).
+            # The destination API rejects null on either field with
+            # "Field may not be null" — strip whichever is null before POST.
+            "attributes.slo_id",
+            "attributes.slo_query",
+            # end is optional (recurring corrections omit it); when the
+            # source resource has end=null the destination API rejects
+            # the payload rather than treating null as "unset". Strip.
+            "attributes.end",
+        ],
         skip_resource_mapping=True,
     )
     # Additional SLOCorrections specific attributes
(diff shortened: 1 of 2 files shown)


riyazsh:
The `Ensure labels` check is failing because a `changelog/*` label is required. Requesting a maintainer to apply `changelog/Fixed` (or an equivalent — this is a defect fix, no new API or removal). I do not have label permissions on this repo via either my personal or EMU account.

michael-richey (approved):

PR #703, opened 2026-09-22
feat: add --alter-logs-indexes-retention-days flag

michael-richey:
## Summary

Add a new `sync`/`migrate` flag, `--alter-logs-indexes-retention-days`, that overrides the standard (non-flex) retention window on logs indexes. It mirrors the existing `--alter-flex-logs-retention-days` flag, which overrides the flex-tier retention window.

## Motivation

`--alter-flex-logs-retention-days` already lets operators override `num_flex_logs_retention_days` on synced logs indexes. There was no equivalent for the standard retention field, `num_retention_days`. This adds the symmetric flag so operators can adjust non-flex retention during sync/migrate without editing imported state by hand.

## Behavior

- New flag: `--alter-logs-indexes-retention-days`, `type=IntRange(min=30)`, `default=None`.
- Registered in `_sync_options`, so it is available on both `sync` and `migrate` (same as the flex flag).
- In `LogsIndexes.pre_resource_action_hook`, when the flag is set and `num_retention_days` is already present on the resource, the field is overwritten with the supplied value. **Insert-if-present semantics**: a resource that lacks the field is never modified — the field is never added, only overridden. This matches the flex flag exactly.
- The two retention flags are independent: setting one does not affect the other's field.

## Implementation

`pre_resource_action_hook` previously had a single hardcoded block for the flex field. Rather than duplicating it for the new field, the hook now drives both overrides from a class-level table:

```python
_RETENTION_OVERRIDES = (
    ("alter_flex_logs_retention_days", "num_flex_logs_retention_days"),
    ("alter_logs_indexes_retention_days", "num_retention_days"),
)

async def pre_resource_action_hook(self, _id, resource: Dict) -> None:
    for config_attr, field in self._RETENTION_OVERRIDES:
        retention_days = getattr(self.config, config_attr)
        if retention_days is not None and field in resource:
            resource[field] = retention_days
```

The flex behavior is byte-for-byte identical to before; future retention overrides are now a one-line table addition.

Config plumbing (`Configuration` dataclass field, `build_config` kwargs extraction, constructor kwarg) follows the same pattern as the flex flag.

## Tests

New file `tests/unit/test_logs_indexes_retention.py` (mirrors `test_logs_indexes_flex_retention.py`):

- CLI acceptance on `sync` and `migrate` (flag value reaches `run_cmd` as `alter_logs_indexes_retention_days`).
- CLI rejection of invalid values (`invalid`, `-1`, `0`, `29` — below the `min=30` floor).
- Hook overrides an existing `num_retention_days`.
- Hook does **not** add `num_retention_days` to a resource lacking it.
- Unset flag (`None`) preserves the existing `num_retention_days`.
- Regression: both retention flags apply independently.
- Regression: the new flag does not touch the flex field.

All values are synthetic (`index-test`, integer day counts).

## Test plan

- [x] `tox -e py311 -- tests/unit/test_logs_indexes_retention.py tests/unit/test_logs_indexes_flex_retention.py` — 20 passed
- [x] `tox -e ruff` — clean
- [x] Full unit suite (`pytest tests/unit`) — no new failures vs. `origin/main` (9 pre-existing subprocess-test failures are unrelated and reproduce on clean `main`)


michael-richey:
Reviewed this end-to-end — nice addition. The new `--alter-logs-indexes-retention-days` flag is wired cleanly and mirrors the existing flex-retention override behavior with the same insert-if-present semantics.

What I validated:
- CLI option registration in `datadog_sync/commands/shared/options.py`.
- Config plumbing through `Configuration` and `build_config` in `datadog_sync/utils/configuration.py`.
- Hook behavior in `datadog_sync/model/logs_indexes.py` after the table-driven refactor.
- New unit tests in `tests/unit/test_logs_indexes_retention.py` (including independence between the two retention flags and no-field insertion guard).
- Local test run: `tests/unit/test_logs_indexes_retention.py` + `tests/unit/test_logs_indexes_flex_retention.py` (20 passed).

Non-blocking suggestion: add a migrate-path invalid-value case (parallel to the sync invalid-value test), since this option is shared across both commands.

michael-richey (commented):
Completed a code + test review; details captured in the PR comment. No blocking issues found.

michael-richey:
Thanks for the review. Addressed the non-blocking suggestion in 12d4fdb0: `test_cli_rejects_invalid_logs_indexes_retention_days` is now parametrized across both `sync` and `migrate` (mirroring `test_cli_accepts_alter_logs_indexes_retention_days`), so the shared `--alter-logs-indexes-retention-days` option is rejected for invalid values on both commands. Local run: `tests/unit/test_logs_indexes_retention.py` + `tests/unit/test_logs_indexes_flex_retention.py` — 24 passed; `ruff` clean.

riyazsh (approved):

